PR #22, opened 2019-08-01
feat: add Icon component for MDX pages. Use a single icon file per icon type

vpicone:
Having 12 different sketch icons and 12 different github icon files is nuts.

I think gatsby-image is overkill for these little guys, we can colocate them into a single component where we'll add icons as needed.

TODO: document adding an icon and using an icon


vpicone:
@emyarod yes, we don't use those anymore. Prob coulda been a separate PR but I was on an unused file tear.

PR #33, opened 2019-08-08
feat: add icon library component

### LINKED ISSUE #7: Icon library birthday cake experience design spec
What is the next progressive step before that will help us reach the full wedding cake experience while still meeting user needs?

- [x]  Review experience roadmap progress. What user stories can be met with the addition of icon metadata and other developments.
https://ibm.box.com/s/wd68kmq15emxfyxc5w4fox7nwec6o309

- [x] Review Wedding cake. What steps can we take to get closer to this experience? https://github.com/carbon-design-system/carbon-website-archive/issues/1094

- [x] Final user story draft for birthday cake experience.

- [x] Design mockup of experience.

vpicone:
closes #7 

Link to icon library page: 
https://carbondesignsystem-git-fork-vpicone-icon-lib.carbon-design-system.now.sh/guidelines/iconography/library

- [x] mobile search bar
- [x] search by descriptors
- [x] filter by category
- [x] group by category and subcategory
- [x] card and category styles
- [x] use definition tooltip?
- [x] replace local metadata.json with dependancy once icon lib 
   (waiting for catgories from https://github.com/carbon-design-system/carbon/pull/3632)


vpicone:
@designertyler added mobile styles based on our conversation

vpicone:
@carbon-design-system/design made my best guess on mobile spec. Wondering if these inputs should go full width or if they should have a margin. I ended up with the margin, cards go full width per IDL implementation.

# with margin
Screen Shot 2019-08-09 at 16 33 48

# without margin
Screen Shot 2019-08-09 at 16 33 57

chrisconnors-ibm:
there is a large "Undefined" category/subcategory in all icons view, and as search results
image
search term "video"
image
and in the Menu:
image

joshblack:
@chrisconnors-ibm FYI the undefined stuff was brought up in Slack and are dependent on the Sketch PR that was just merged in.

chrisconnors-ibm:
The menu contents order should be alphabetical (Actions, Brand, Enterprise, Organization, Person, Planning, Tools) to match the page layout order.
image

chrisconnors-ibm:
It looks like, for example, all of `Brand` subcat `Social` is appearing as `undefined`.

jeanservaas:
real quick, i know this is left field... i just happened in here to check out the progress... but can we get rid of the rules when there's a hover ... I noticed it in the example in Connors's comment above so I went and looked. 

image

designertyler:
@jeanservaas it looks like that issue is in the React component. I'll create a separate issue for it.

http://react.carbondesignsystem.com/?path=/story/dropdown--default
 
image

jeanservaas:
@designertyler yes, really weird, at first i didn't think it was in the React component, but the rule comes and goes when you move your cursor. 

One other thing to note though, the right side of the dropshadow on the drop down is also cut off and this doesn't not appear to be a problem in the React component.

vpicone:
@chrisconnors-ibm alphabetized fixed
@jeanservaas shadow stuff fixed

@joshblack I believe the issue was due to importing `*` from `@carbon/icons-react` while having two versions of the module (`10.4` from carbon-elements which is pinned and `10.5-alpha-` so I could have the 32px version and latest categories.) 

Using @latest for the icons packages in the latest commit solved the issue.

jeanservaas:
noice, shadow LGT little old me

chrisconnors-ibm:
Testers: refer to https://github.com/carbon-design-system/carbon/blob/master/packages/icons/metadata.yml

search should hit on: `name`, `friendly name`, `aliases`, `category`, `subcategory` values from  `metadata.yml`

vpicone:
@chrisconnors-ibm search now hits on name and category in addition to friendly name, aliases, and subcategory.

designertyler:
Subcategories should be shown in alphabetical order.

There are 16 broken icons like this. I'm guessing this will be "fixed upstream"?
image


Something is off with the clear X when there's a term entered in the search field. This is on hover. 
image


The dropdown is limited to 3.5 rows. Can we fully extend it?
image


If there's a serch term that has results in another category but not the selected one, the "no results" state should link to "all icons" like the mockup here https://github.com/carbon-design-system/carbon-www/issues/7#issuecomment-516041842

image

src/components/IconLibrary/IconCategory.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { groupBy } from 'lodash';
```
joshblack: Super unrelated to the PR, but do we have something like `babel-plugin-lodash` setup for `lodash` for bundle size stuff? If not, should we just use the package directly? e.g. https://www.npmjs.com/package/lodash.groupby

src/components/IconLibrary/IconLibrary.module.scss
```diff
@@ -0,0 +1,199 @@
...
+
+// i.e. fallback
+.icon-grid {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.icon-grid {
+  margin-top: $spacing-05;
+  display: grid;
```
joshblack: Awesome to see CSS Grid, is this okay for browser support? If so then 💯

src/components/IconLibrary/IconLibrary.module.scss
```diff
@@ -0,0 +1,199 @@
...
+
+  @include carbon--breakpoint('lg') {
+    grid-template-columns: repeat(6, 1fr);
+  }
+}
+
+// covers the excess space at the end of a grid
+.icon-grid::after {
+  content: '';
+  width: 600%;
```
joshblack: This rule was surprising to see with the `600%`, could you talk through it a bit more?

src/components/IconLibrary/iconMetaData.js
```diff
@@ -0,0 +1,28 @@
+import { keyBy } from 'lodash';
+import { icons as iconMetaData } from '@carbon/icons/metadata.json';
+import { pascal } from 'change-case';
+
+const flattenedIconMetaData = iconMetaData.flatMap(
```
joshblack: You had said before that it'd be super helpful for us to expose a flattened list of icons right? Similar to what's in build-info? Trying to figure out what I can fit into the icon-cli stuff so that we can offer this stuff for anything doing icon work 🤔

src/components/IconLibrary/IconLibrary.module.scss
vpicone: I have a fallback right above.

src/components/IconLibrary/iconMetaData.js
vpicone: Yeah the flattened icon list and the prefix/module name would have been incredibly helpful in the implementation here

vpicone:
@joshblack not sure if you saw my note above but re: bundles I’m pretty sure that was fixed yesterday, let me know if you still think that’s not the case. 

> I believe the issue was due to importing * from @carbon/icons-react while having two versions of the module (10.4 from carbon-elements which is pinned and 10.5-alpha- so I could have the 32px version and latest categories.)

src/components/IconLibrary/IconLibrary.module.scss
vpicone: Initially I just had a grid gap with a background color underneath to avoid the weird nth rule stuff. This covered the worst case scenario of a single orphan icon by spanning 6 items' width. Unfortunately, the 1px grid gap had some weird sub pixel stuff going on.

This should be taken out along with the container background color.

vpicone:
@designertyler 
- yeah those are upstream
- fixed the close button
- updated dropdown item background and max height
- added `All icons` link with result count when the category filter is preventing icons from being shown

	For the "0 matches for X in Y" text, it looks really weird to have both the text input and the `X` text updating together as the user types. I think that sort of tagline is more for a request that hits a server as opposed to something that updates in real time. I think the big `all icons` link and missing result count gets the point across that the category is what's causing there to be no results.

joshblack:
@vpicone how'd you verify that it was resolved?

vpicone:
@joshblack sorry I thought you were talking about the duplicate icon bundle that was happening from using the canary with the regular icons package.

The icons get put in the app bundle because they're used on other pages/components so the resources are shared. The current carbon website makes two requests for the same data when you load the home page. Icons are also included in the regular gatsby theme app.js at the moment.

joshblack:
> The icons get put in the app bundle because they're used on other pages/components so the resources are shared. 

I would assume only a subset of icons are ultimately shared, right? Surprised that it doesn't bring those chunks in and defer the reset till later. (Although that sounds gnarly and I do not envy that challenge on the bundler side lol)

As a part of this effort it'd be great if we could address this as it can lead to a poor user experience. Is there a performance budget for the site we could measure against? When looking at webpack, it seems like they suggest asset chunks stay <250kB, would that be something we'd want to mirror?

vpicone:
@joshblack proper dynamic imports is a possibility with the icons module name/prefix. At the moment I have to import the library in its entirety, I imagine that could be a vector for performance improvements down the line once we have the build metadata.

I'm not sure why the icons got put into a separate chunk for the old website there's a lot of unknowns. However, it does load the icon chunk in its entirety on the home page load as well. Maybe it has something to do with the lack of a default export on the icons-react package?

If you want I can keep exploring trying to chunk the icons onto their own.

joshblack:
> At the moment I have to import the library in its entirety, I imagine that could be a vector for performance improvements down the line once we have the build metadata.

Yeah, definitely understand that. Something we could do is publish `build-info.json` and use that to derive the direct paths as it has a flat list of icons and their module paths (so webpack could in theory chunk by file instead of one glob)

When it comes to your last line, I think we should definitely try and figure out why our bundle size effectively doubled (329kB on live site to 705kB on preview URL, 1.9MB uncompressed to 5.8MB uncompressed). Would you agree? Or is this unrelated to changes made in this PR?

designertyler:
After clicking the all icons link the category selection should change to all icons. Right now I click the link, it takes me to all icons, but the category doesn't update.

Screen Shot 2019-08-14 at 11 36 25 AM

> For the "0 matches for X in Y" text, it looks really weird to have both the text input and the X text updating together as the user types.

Fair enough. Was it intentional to leave these parts? It's not really valid if there are results in another category. 

Screen Shot 2019-08-14 at 11 50 54 AM


At narrow widths, the resource cards expand all the way into the right margin 
image

Here's how the cards should behave with the margin.

image

vpicone:
@designertyler I thought it still made sense because if they're looking for an icon X in category Y and it doesn't exist then they'd still want to contribute. 

It sort of depends on whether we assume they are accidentally narrowing their search (whoops, go back to all icons). 
Or we assume they're being intentional with their search: "Oh, there's no `Health` icon in the `Brand` category, let me contribute one!"

designertyler:
The scenario of accidentally narrowing search is what we were solving for. It needs to be clear that the icon they searched for (`health`) does exist even if it's not in the category they expected (`brand`).

That's why we had the text for "0 matches found in the category you selected... but hey, we do if you expand the category to all." There's no need to distract with the concept of submitting an icon if we **do** have icons that match their search term.

vpicone:
Totally, I was thinking if they have an _intentionally_ narrow search i.e. looking for a Watson Health brand logo. If it didn't exist, we'd want them to contribute it. I'll take a crack at the design as intended now.

vpicone:
@joshblack the bundle size doubled because of the icons, unfortunately the dynamic imports of just the size we need isn't possible because of the lack of metadata.

The carbon design system website benefits from not receiving 4 assets for every icon at the moment. It also splits the icons into a separate chunk, but that chunk is still loaded on the homepage. Even with these two factors, the overall content sent over the wire is comparable between this PR and the carbon design system website (which clocks in around 800kb atm).

vpicone:
@designertyler 
- added a `0 result found in Category` string. 
- removed contribution resources when icons exist in all categories
- fixed margin at tablet width
- fixed category reset

joshblack:
Makes sense @vpicone! Thanks for following up on this, really appreciate it 🙏 

Interacting with the page, was curious if there is supposed to be a tooltip on the icon names? Unfortunately it isn't triggering and it looked like the titles had a cursor style on them 🤔

vpicone:
@joshblack tool tip wasn't working in chrome for some reason, it's good now.

vpicone:
@laurenmrice should be fixed now thank you!

vpicone:
@laurenmrice after some digging it looks like IBM Firefox is no longer a thing: https://w3.ibm.com/help/#/article/mozilla_faqs

That version of firefox (60) is no longer receiving updates, so it'll continue to lag behind the current (68).

designertyler:
Can confirm it looks fine in Firefox 68.

Tooltip should only be enabled on truncated text.

vpicone:
@designertyler the tooltip is an all or nothing thing. Since we truncate the text with css, there’s no way for us to know if a given label is truncated or not at a given screen width. Sometimes it’s cut off between media queries.

designertyler:
@vpicone Understood. I'm discussing the possibility of changing the spec to wrap the text to two lines rather than truncate at one line. FYI @jeanservaas @shixiedesign @laurenmrice 

There's also an issue with the tooltip being cropped with longer titles on the right side.

image



### Truncated (as-is)
Screen Shot 2019-08-19 at 1 53 53 PM





### Wrapped text
Screen Shot 2019-08-19 at 1 50 232

shixiedesign:
Personally I think it looks totally fine with 2 lines for name. I definitely don't want to go to 3-lines though. Is it possible to put a character count limit on the names for current (and contribution) icons.

laurenmrice:
yep lets do two line wrap

mjabbink:
current IDL has two lines in  one spot: "Checkbox undeterminate"
https://www.ibm.com/design/language/elements/icon-library

I think it will fine too as long as it’s not every icon!

mjabbink:
I assume the tool tip is only for truncated names? Why else would it be there?

laurenmrice:
The newest build got rid of the tooltips and now wraps the icon name text if it runs out of room.

mjabbink:
that is probably better solution since having unneeded tool tips is not great.

mjabbink:
The search and category dropdown should be sticky

vpicone:
@mjabbink I'll add the sticky search/dropdown

Only adding tooltips to truncated names is tough here since we need to check every name when the page loads and again whenever the browser changes sizes. That sort of thing works on say table headers (5-10 items), but becomes a huge performance hit on the scale here (700+ icons)

vpicone:
@mjabbink are you okay with putting the sticky search in a new issue since we don't have a spec for that at the moment?

mjabbink:
I don’t think we need the “Categories” label over the dropdown. I think it should be removed

designertyler:
@vpicone @mjabbink Are we fine moving the sticky search to another PR after specs are done?

This PR should be ready to be approved once the titles wrap (glad to see IDL doing it!) and the "categories" label is removed.

jeanservaas:
@vpicone @designertyler we are definitely good to put sticky search in another PR — it's not ready, we need a full spec for it

chrisconnors-ibm:
@designertyler Please consolidate sticky search controls and any necessary iteration of search controls into a separate issue and link it here.

PR #35, opened 2019-08-09
fix: missing styles

alisonjoseph:


src/styles/_page.scss
```diff
@@ -0,0 +1,78 @@
...
+  @include carbon--type-style('expressive-heading-05', true);
+}
+
+.page-h2 {
+  margin-top: $spacing-11;
+  margin-bottom: $spacing-05;
+  @include carbon--type-style('expressive-heading-04', true);
+}
+
+.page-h3 {
```
alisonjoseph: @vpicone is there another way to pull in these styles without having to duplicate them??

vpicone:
```jsx
import {
  h2,
  h3,
} from 'gatsby-theme-carbon/src/components/markdown/Markdown.module.scss';

import {
  iconCategory,
  categoryTitle
} from './IconLibrary.module.scss';

// then 
 return (
    <section className={iconCategory}>
      <h2 className={cx(h2, categoryTitle)}>{category}</h2>
```

is that what you meant?

alisonjoseph: Yea, that should work I think. thx!

PR #55, opened 2019-08-15
fix: add inverse icons to homepage

### LINKED ISSUE #52: github icons on homepage need to be white
Screen Shot 2019-08-14 at 6 34 29 PM

alisonjoseph:
closes #52


vpicone:
Maybe we should just have an `inverse` option since `inverse-01` might not necessarily be white all the time.

alisonjoseph:
good call, I updated

PR #112, opened 2019-08-20
feat: update header height

alisonjoseph:


vpicone:
This might break sticky tabs.

alisonjoseph:
@vpicone yep it does. 😭

alisonjoseph:
sticky header is fixed.... but this is a temp fix. we need to add this as an option back to the theme because the sticky stuff at mobile hides "late" now on Carbon.

src/styles/_page.scss
```diff
@@ -1,3 +1,9 @@
+// Override for header height, possible theme option in the future
+body [class^='PageHeader-module--page-header'] {
+  height: rem(240px);
+  top: -9rem;
+}
+
```
vpicone: Need to update how much the page header get's shifted so we don't shift the tabs off the screen.
```suggestion

[class^='PageHeader-module--page-header--shifted'] {
	top: 12rem;
}
```

vpicone:
I'm gonna update the tabs to respond to whatever the page header height in this pr https://github.com/carbon-design-system/gatsby-theme-carbon/pull/364. 

Now that we don't have to worry about sticky header, it makes it easier to handle in one spot.

PR #118, opened 2019-08-21
chore: update carbon packages

alisonjoseph:


jnm2377:
A few issues I ran into, I know you said some of these are known:

- component docs `html` formatting
- guidance > colors > usage page: when you switch color tabs the page goes completely black for me. If I refresh you can see the page again, but it doesn't switch color tabs. 

bug

alisonjoseph:
@jnm2377 thanks so much for reviewing! I opened an issue to look into the Color page. the component pages should be formatted correctly now.

PR #125, opened 2019-08-21
docs: Experimental overview page refresh

connor-leech:
* Separate usage tab was unnecessary and poorly placed. Feature flag info now lives under "Overview".
* Left nav now matches header


jeanservaas:
Something weird with the anchor link spacing:

image

Here it is on the live site:

image

PR #130, opened 2019-08-22
fix vue tutorial step 5: reference cf instead of ibmcloud in Create m…

ZrianinaMariia:
fix vue tutorial step 5: reference cf instead of ibmcloud in Create manifest file for consistency, since it is used below this section.

Changes
- src/pages/tutorial/vue/step-5.mdx


ZrianinaMariia:
> Thanks for spotting this Mariia,
> 
> could you also address the three uses of ibmcloud in the react version.

I could not find references to `ibmcloud` in the react version. It was `cf` everywhere, that's where I took the link to cf installation instructions. I believe the current website version does not reflect the `master` branch of this repo.

Issue #245, opened 2019-09-05
docs: Improvements to the Content > Glossary


tomwaterton:
## Background
After discussions with @janchild and @connor-leech , I have spent some time reviewing the existing Carbon > Content > Glossary page (as well as some other similar glossary / terminology sets at IBM) and have a few proposals. (This issue does not cover the actual terms listed themselves. I'll do that separately.)

Note: I'm happy to do this work and submit the PR, but thought I'd raise this in an issue first, so that others have a chance to comment on these proposals.

### 1) Add an intro sentence or two to explain the intended scope

Rationale: 
- I imagine (and hope) that if we invest some time and effort in updating and enhancing this Carbon Glossary, more and more teams across IBM might start using it. 
- Indeed, if we do a good job, I'd hope that over time people might decide they no longer need to maintain their own local glossaries. 
- However, if more teams use this glossary, we'll no doubt have more people asking us to add more terms. This is all fine - but I think it would be wise to set expectations now re what the scope of the glossary is. 
- I _believe_ that the Carbon Glossary is intended as a lightweight resource for designers and engineers to cover the most common terms that get used across many IBM product UIs (e.g. common action labels). It is therefore not intended to cover endless product-specific technical terms and concepts which might only apply to one or two products. 

So, I propose we add some text to the top of the page. Here are some examples: 

> _The Carbon glossary is intended as a lightweight reference, covering common UI terms._
or
> _The Carbon glossary is intended as a lightweight reference covering terms commonly used in many IBM product UIs. Product-specific terminology should be covered elsewhere, such as in product-specific docs._
or 
> _The Carbon glossary is intended as a lightweight reference covering common UI terms. It’s not intended to cover general computing concepts nor product-specific terminology, which should be covered elsewhere, such as in product-specific docs._  

Thoughts?

### 2. List terms in lowercase
- Carbon guidance is to use sentence case in our UIs (so we don't want people to think that they need to always capitalize these words).
- To support this, I think it would be more helpful to list terms in lowercase by default (as is common in dictionaries, etc.) 
- The only exceptions would be:
  - proper nouns (e.g. Watson)
  - acronyms (e.g. ASCII)
  - initialisms (e.g. IBM, HTML, etc.)

### 3. Remove all instances of the sub-heading "Action labels"
I believe that under each letter section, originally there were two sub-headings: 
- Action labels
- Idioms  

[...]

PR #252, opened 2019-09-05
fix: remove unnecessary entry title

vpicone:
Addresses issue number 3 in #245


connor-leech:
Thanks for doing this so quickly 🙌

PR #341, opened 2019-09-18
Split component example

asudoh:
This change extracts out instantiation/release code of vanilla JavaScript library from `<ComponentExample>` and plain-HTML-specific code from `<CodeExample>`. Doing so paves the cowpath toward better integration of React live demo to those component.

Refs carbon-design-system/carbon#1936.

#### Changelog

**Changed**

- Extracted out vanilla component instantiation code to a separate module/component. (`<ComponentExampleLive>`)
- Changed `hasLightVersion` prop of `<ComponentExample>` reflecting the usage in code.
- Extracted out the code to load HTML for components (variants) to a separate module. (`getHTMLFile()`)
- Enhanced `<CodeExample>` with a choice of language (notably supporting JSX).
- Made Prism usage in `<CodeExample>` more ESM-oriented.

**Removed**

- `experimental` prop from `<ComponentExample>` since it's no longer in use.
- Local `componentNameLink` variable from `<ComponentExample>` rendering code since it's no longer in use.


asudoh:
Thank for reviewing and good catch @alisonjoseph! Fixed the `field-0x` switcher.

PR #353, opened 2019-09-19
chore(ComponentExample): support React live demo

asudoh:
This change adds React live demo and code snippet feature to `<ComponentExample>`.

* The React live demo feature is implemented as `<ComponentExampleLiveReact>`, which has the same API as its vanilla conterpart (`<ComponentExampleLive>`) and thus can be swapped back and forth by user's selection of JavaScript framework variants for demos
* The React code snippet feature is implemented as `<CodeExampleReact>`, which has the same API as its plain HTML counterpart (`<CodeExampleHTML>`) and thus can be swapped back and forth by user's selection of JavaScript framework variants for demos

The React live demo feature is enabled with `hasReactVersion="live"` prop of `<ComponentCode>`. It can be also set as the default by `defaultFrameworkVariant="react"` prop of `<ComponentCode>`. If `defaultFrameworkVariant="react"` is set, vanilla live demo won't be enabled unless `hasVanillaVersion={true}` prop of `<ComponentCode>` is set.

The source of the live demo as well as of the code snippet comes from a map of React components, keyed by the component name and the variant name specified for `<ComponentCode>`, which is located at `src/react-demo/index.js`.

Special note for code snippet: The code snippet for the React version is generated as a conversion from a shallow React component tree to a JSX string. Due to its "shallow" nature, using `<DemoComponent foo={fooValue}>` for the React-component-tree-to-JSX converter just emits `<DemoComponent foo={fooValue}>` (as the code snippet), and thus we run `DemoComponent({ foo: fooValue })` instead to pass along to the converter to get the JSX of the React component tree of what `DemoComponent` generates. Those demo components has to be implemented as pure functions therefore.

This change also removes drop-up demo for now due to the limited availability of the demo.

Refs carbon-design-system/carbon#1936.

@cal-smith Updated Angular links to point to the right Storybook page, but please don't hesitate speak up any links are broken in the staging page. Thanks!

#### Changelog

**New**

- `<ComponentExampleLiveReact>` a component that achieves React live demo and replaces `<ComponentReact>`
- `<ComponentCodeReact>` a component working as code snippet for React

**Changed**

- Dropdown, multi select and combo box page to use the new React live demo feature (and use it as default)

**Removed**

- `<ComponentReact>`, superseded by `<ComponentExampleLiveReact>`


alisonjoseph:
exciting!!!

alisonjoseph:
This is great! So excited we're finally getting this added into the website 😸🎉

Couple questions..

The demo code for dropdown for React is showing
```
<t
      ariaLabel="Dropdown"
      helperText="Optional helper text"
      id="dropdown-9-0"
      itemToString={function(e){return e?e.text:""}}
      items={[ { id: 'item-1', text: 'Item 1' }, { id: 'item-2', text: 'Item 2' }, { id: 'item-3', text: 'Item 3' }, { id: 'item-4', text: 'Item 4' } ]}
      label="Dropdown menu options"
      titleText="Dropdown label"
    />
```

Is that correct or should it be showing `<Dropdown .... `> ?

Also, it feels a bit off when you click on vanilla/react to have them swap places, I'd suggest we keep them in the same place and just swap from a link to a label? Aslo tagging @aagonzales since I know she's been working on some different updates to this page to see if she has an opinion on how the react/vanilla code switcher should work. 
Screen Shot 2019-09-20 at 7 09 24 AM
Screen Shot 2019-09-20 at 7 09 20 AM


Also something for a future PR maybe? Have you seen the live code option for MDX?  https://mdxjs.com/guides/live-code

aagonzales:
Yes this is good to see! I like the functionality but I think we'll want to change the visuals a little so it looks more like tabs or a content switcher than a link out. I'll sketch up some ideas. 

**Questions:**
- Will every component eventually get this react code tab as well?
- I'm also curious to know if dev wants to keep showing the code on the page like this or if we just want to link out to all the repo/dev environments and not show code on the website.

asudoh:
Thanks everyone for the feedback. Pushed a quick fix for minified component name. Looking forward to seeing the sketch from @aagonzales! Wrt the live code stuff, something that sounds very cool (thanks @alisonjoseph for the idea!), OTOH I'm also with @aagonzales that code snippet may be more of dev space. Great to have more discussion within the team there.

Lastly, though this PR is of showing correct dropdown selection by default in website (as part of addressing carbon-design-system/carbon#1936) great if the website team can allocate time to go through other components as the team and apply the same change. Thanks again!

src/components/CodeExample/CodeExampleReact.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from 'prop-types';
+import reactElementToJSXString from 'react-element-to-jsx-string';
+import CodeExample from './CodeExample';
+import demos from '../../react-demo';
+
+let instanceSeq = 0;
+const getInstanceId = component => `${component}-${instanceSeq++}`; // eslint-disable-line no-plusplus
+
+const CodeExampleReact = ({ component, variation, useLightVersion }) => {
+  const instanceId = useMemo(() => getInstanceId(component), [
```
vpicone: This should be stored in a ref, not memoized

aagonzales:
Here's my ideas for updating this now (before the full component live demo updates happen). Adding tabs seemed like too much with the tabs on the page. I like option 1 the best, seems cleanest. 

### Option 1
Add a inline dropdown to switch the language 
image

### Option 2
Add a content switcher 
image

alisonjoseph:
I also like version 1 with the select. 

Would we want to keep the vanilla link and have it link here? https://the-carbon-components.netlify.com/ We don't currently link out to anything other than the codesandboxes right now,.

asudoh:
Thanks @aagonzales for coming up with UI options! Went with the option with inline select given it seems our consensus. Introduced link-out feature for vanilla, but it can be removed if it's more desirable. Given this PR focuses on getting React example on-par with vanilla, let's separate the discussion of live edit of code snippet of the demo.

vpicone:
If possible we should avoid having a directory of code samples that get pulled in dynamically. The advantage of using MDX is using the react components in the document. This abstraction for vanilla ha made it hard for folks to maintain examples.

We do this for the gatsby theme documentation and it's been extremely successful for maintaining code snippets and usage guidelines.

asudoh:
Removal of React code example is fine, as long as other people agrees. We seem to need another design in such case, given the current design tightly integrates the select box to choose React/vanilla in the area of code example.

alisonjoseph:
I agree, will be another entire thing to maintain if we have to create a new React file for each component. 🤔

asudoh:
@aagonzales Would you think of a quick UI sketch without the code snippet portion...? Thanks!

alisonjoseph:
What is the goal of this PR? Is it just to update Dropdown examples so we can show the correct variant? https://www.carbondesignsystem.com/components/dropdown/code

Trying to do anything more, we should wait until design in #357 is ready. (keeping in mind we should be building this with React in mind)

aagonzales:
I agree with Alison we should hold off on making any major changes until #357 is ready. For now, we should just show the react only versions the same why they currently are being showed on the website (image shared below). 

(side note: if y'all have opinions about what we should do with the code blocks in the future please post in the #357 issue). 

image

vpicone:
@aagonzales #357 is going to be my next project think we should focus efforts on that than rather than put the time into creating examples for all these components only to have them taken away when we replace with the live view.

aagonzales:
Yes but we still need to showcase this new react only component on the website before the new project is done. And the easiest/lowest cost is to do it like we already are doing it on the [dropdown](https://www.carbondesignsystem.com/components/dropdown/code/) page. The sample window but no code block attached to it. 

This is how we should handle this now as is imo
image

asudoh:
@alisonjoseph The primary goal of this PR is showing React version for drop down, given it has the correct UI implementation with regard to selection.

@aagonzales What do you think making our (regular) dropdown showcase like multi select one? I have an impression from your latest screenshot that it may be OK (my apologies otherwise).

vpicone:
@asudoh if all we need is to show the component in action, let's just use it in MDX. No need for any of the abstraction/CodeExample stuff.

asudoh:
@aagonzales Wanted to double-check if the main dropdown example should look like multi-select one? Thanks!

aagonzales:
@asudoh yes, I think that's best for now.

asudoh:
Thanks @aagonzales for your response! Superseded by: #421

PR #370, opened 2019-09-21
refactor: remove vimeo in favor of native video element

vpicone:
Removes Vimeo library and styles in favor of native video element. Decreases bundle size and request count.


src/components/HomepageVideo/ToggleIcon.js
```diff
@@ -0,0 +1,57 @@
...
+        height="32"
+      />
+    </svg>
+  ) : (
+    <svg
+      className="mouseOut"
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 32 32">
+      <path d="M16 4A12 12 0 1 1 4 16 12 12 0 0 1 16 4m0-2a14 14 0 1 0 14 14A14 14 0 0 0 16 2z" />
+      <path d="M12 23a1 1 0 0 1-.51-.14A1 1 0 0 1 11 22V10a1 1 0 0 1 .49-.86 1 1 0 0 1 1 0l11 6a1 1 0 0 1 0 1.76l-11 6a1 1 0 0 1-.49.1zm1-11.32v8.64L20.91 16z" />
```
abbeyhrt: Looking at the svg paths, I can't tell if these are Carbon icons. Are they Carbon icons?

vpicone: I'm not sure, just kept the ones that were there, wonder if we should just switch to use icons-react

alisonjoseph: I think this might have been built before the icons-react package was being used on the website, so we could probably switch to use play-outline/filled and pause-outline/filled

abbeyhrt:
The play/pause button is showing the play button when it is playing and the pause button when it is paused for me, it should be switched right?

vpicone:
@abbeyhrt why you starting trouble

abbeyhrt:
lol I'm sorry!

vpicone:
@abbeyhrt should be good to go

PR #441, opened 2019-10-04
fix: add classname prefix to component example wrapper

alisonjoseph:
Fix classname conflict coming from carbon-charts that is affecting `tooltip` demo.

Screen Shot 2019-10-04 at 10 44 55 AM

#### Changelog

**Changed**

Added `component-example--` to all component wrapper class names

#### Testing

Check component demos, specifically tooltip, button and loading to make sure they are all working as expected.


alisonjoseph:
Seems like the definition tooltip styles are being overridden by the reset. I'm not 100% sure what's happening there. (I have a feeling this was broken before the charts stuff got merged in) 

🤔

laurenmrice:
- The interactive tooltip (middle example) looks correct when I preview it based off of what we decided on in this pr: https://github.com/carbon-design-system/carbon/pull/3476
- Yep, the Definition tooltip examples should not all be center aligned.

emyarod:
oh I was basing the middle interactive tooltip example off of the current netlify preview which looks like this

image

alisonjoseph:
@emyarod the html is different, we are pulling it in from carbon-components `10.6.4` where the netlify link is just latest in master I believe.

asudoh:
Wrt tooltip reset thing - Any change for us to pull in #4247? (Patch release or a like?)

alisonjoseph:
I don't believe we can pull in anymore fixes while in code freeze, is that correct @jnm2377

alisonjoseph:
Yea, I think we can merge this then, it fixes the original issue brought in my adding charts, and then add the patch fix once ready.

alisonjoseph:
I opened a new issue https://github.com/carbon-design-system/carbon-website/issues/467

alisonjoseph:
@laurenmrice are you ok w/ merging this and addressing the definition tooltip in a new PR?

PR #457, opened 2019-10-07
fix(notification): style tab updates

aagonzales:
Updating information on notification/style/color section to include high and low contrast content. Also restructured color section so it was easier to consume/process. 

(updating usage content in separate PR)


aagonzales:
Not currently in the code. It has the drop-shadow. We need to do some follow up on the code side of things about notification styling. 
https://the-carbon-components.netlify.com/?nav=notification

PR #475, opened 2019-10-09
fix: temporarily disable lock file to pin charts

vpicone:
Latest charts release is broken: https://github.com/carbon-design-system/carbon-charts/issues/363

Because their release process is also broken, we can't lock those packages down, yarn ignores their entries.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,22 +1,21 @@
 name: Test site build
 
-on: 
+on:
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


theiliad:
Fixed in v`0.16.15`. Also tried the build on the website with the latest version of charts, it's working well

image

image

theiliad:
Fixed in v`0.16.15`. Also tried the build on the website with the latest version of charts, it's working well

image

image

Issue #491, opened 2019-10-15
Add a11y documention to component pages


alisonjoseph:
Add Accessibility tab and documentation for each components page on carbondesignsystem.com

- [X] Accordion
- [X] Breadcrumb
- [ ] Button
- [X] Checkbox
- [ ] Code snippet
- [X] Content Switcher
- [X] Data table
- [X] Date picker
- [X] Dropdown
- [X] File uploader 
- [X] Form 
- [X] Inline loading 
- [X] Link 
- [X] List
- [X] Loading
- [X] Modal 
- [X] Notification
- [X] Number input
- [ ] Overflow menu
- [X] Pagination
- [X] Progress indicator 
- [X] Radio button 
- [X] Search
- [X] Select
- [X] Slider 
- [X] Structured list
- [X] Tabs 
- [X] Tag
- [X] Text input
- [X] Tile
- [X] Toggle
- [X] Tooltip
- [X] UI Shell

https://pages.github.ibm.com/IBMa/Carbon_Integration/

snidersd:
I noticed there are 3 components that do not have an accessibility tab, just wondering when they are expected to go live. Thx!
- [Button](https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/button-doc.html) 
- [Overflow Menu](https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/OverflowMenu-doc.html)
- [Code Snippet](https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/CodeSnippet-doc.html)

snidersd:
Please update the link to WCAG in all the documents from:
[WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) 
To: 
[WCAG 2.1 AA](https://www.w3.org/TR/WCAG21/)

PR #494, opened 2019-10-16
feat: add accordion a11y docs

alisonjoseph:
Ref #491 

### New

Accessibility tab on Accordion component page

### Testing

Content should match https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/Accordion-doc.html


jeanservaas:
Just quick comment about the hierarchy and tables:

Tables aren't maybe the most ideal for all of these stepped test results, but until we have a content and layout pass, this will have to do. 

**- let's make the tables looks consistent with the rest of the site with the 48px header bar**

image

**- you can use and H4 for the titles of the tables as long as you can get the 48px margin top and 24px margin bottom... if not use the H3 if it's easier**

**- if there are only two columns within the tables, the second column needs to align**

alisonjoseph:
I don't think there is anywhere on the website that has that spacing, not sure we want to add some weird spacing overrides just for this page? 

this is h2 + h4 and adding spacing above the structured list
Screen Shot 2019-10-24 at 3 58 36 PM

this is h2 + h3, which doesn't look great, but its what we have everywhere right now
Screen Shot 2019-10-24 at 3 59 42 PM

jeanservaas:
let's do the H2 with the H3 then.

src/pages/components/accordion/accessibility.mdx
```diff
@@ -0,0 +1,382 @@
...
+
+1. Each accordion header must have a unique title (implemented via the label for the button) that clearly describes the accordion panel content. This is particularly helpful for users of assistive technologies so they have the necessary information to efficiently navigate the content without having to expand every section.
+1. Carbon Components should be used to create the content that displays within each accordion panel.
+1. Avoid keyboard traps when adding components to the accordion panel. For example, the user expands an accordion, but is unable to tab to the next focusable element.
+1. Although the accordion component passes accessibility testing, content authors need to ensure the content that is added to the accordion is accessible. For example, if you add an image to the accordion header or panel you need to include alternative text to pass accessibility testing.
```
aledavila: should this be 1-4

aledavila: should this be # 1 - 4

alisonjoseph: It could be, but actually doesn't matter, it will get converted to `<ol>` no matter what the numbers actually are.

aledavila:
Some of the docs were updated for accordion. Do you want me to push the updates or do you want to do it. Its some changes to environment.

alisonjoseph:
Sure go ahead

src/styles/_structured-list.scss
```diff
@@ -0,0 +1,29 @@
...
+// component example
+//
+//---------------------------------------
+.bx--row > * > .bx--structured-list .bx--structured-list-row {
+  display: flex;
+  justify-content: space-evenly;
+}
+
+.bx--row > * > .bx--structured-list .bx--structured-list-row--header-row {
+  background: $carbon--white-0;
```
alisonjoseph:
```suggestion
  background: $ui-background;
```

PR #506, opened 2019-10-18
Update $danger for g90 and g100

oliviaflory:
All $danger values should be Red 60 across all themes.

#### Changelog

**Changed**

- changed values for $danger in g90 and g100


aagonzales:
Waiting on @shixiedesign to give us the official change log file to verify this is correct with the v2 color updates.

PR #533, opened 2019-10-28
Better build time

vpicone:
This uses a gatsby static query rather than running a shell script and using raw loader.


alisonjoseph:
Thanks for the fix! should we update IDL also?

vpicone:
@alisonjoseph opened here: https://github.com/carbon-design-system/design-language-website/pull/332

PR #563, opened 2019-11-04
Docs: Fixed Accordion-usage-1.png

chloepoulterdesign:
We spotted that the accordion chevron was upside-down in the accordion-usage-1.png image, so I've updated it.


chloepoulterdesign:
Hi @vikkipaterson could you please approve this small image change on the Accordion component?

vikkipaterson:
Done

PR #588, opened 2019-11-14
docs(tag): add a11y tag docs

aledavila:
Reference #491 

Adds a11y tag docs to component page

Make sure content matches: https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/Tag-doc.html


snidersd:
The Resources section has a subtitle
 "Helpful resources for creating customized accessible implementations" 
which was the section title in the original documents. It is present in the Tag document and missing from others I've reviewed (i.e. Accordion & Breadcrumb). My recommendation is to make sure that all the titles and subtitles are the same for consistency. Thx

snidersd:
Is it too late to change the DAP results to No violations now that the DAP issues fixed?

PR #592, opened 2019-11-15
fix(notification): images updated

aagonzales:
Updated low contrast notification images with correct/updated style.


aagonzales:
Hmm I don't know what the problem is with the merge conflict.

PR #596, opened 2019-11-18
chore: lean on theme version of certain deps

vpicone:


vpicone:
Looks like this makes a pretty significant difference in bundle size, worth exploring in gatsby theme v2

PR #599, opened 2019-11-18
fix: add overflow container for tables

jnm2377:
Closes https://github.com/carbon-design-system/gatsby-theme-carbon/issues/496

Carbon's component status page and data table component demos do not have responsive scrolling. The issue was opened in Gatsby theme repo, but this is actually a carbon bug since in order to make tables horizontally scroll, they must be wrapped in a scrolling container, which has to be implemented on the carbon  website side (not something the theme can control).

#### Changelog

**Changed**

- component demo styling: added overflow-x scroll, and tweaked some other styles so that it will be responsive in mobile breakpoints.


src/styles/_overrides.scss
```diff
@@ -159,3 +159,13 @@
   margin-left: 1.375rem;
   margin-top: 2.5rem;
 }
+
+//---------------------------------------
+// Data table
+//---------------------------------------
+
+.component-example__live--data-table {
```
vpicone: I don't think we need this or the TableContainer component anymore right?

jnm2377: Ah I meant to delete this. But the TableContainer component is being used in the `ComponentStatus` @vpicone

src/components/ComponentStatus/ComponentStatus.js
```diff
@@ -56,22 +57,24 @@ class ComponentStatus extends React.Component {
...
+                  <th>Vue</th>
+                </tr>
+              </thead>
+              <tbody>
+                {Object.keys(componentStatus.components).map(component =>
+                  this.renderItems(componentStatus.components[component])
+                )}
+              </tbody>
+            </table>
+          </TableContainer>
```
jnm2377: @vpicone this is where the TableContainer component is being used

vpicone: Ah I see. We might be able to just style ComponentStatus rather than creating a component. 

We could make a ComponentStatus.module.scss and style it there.

jnm2377:
@vpicone updated

PR #619, opened 2019-11-22
docs(react): improve .cfignore

docwhat:
With a proper `.cfignore` the deploy speed is about 25 times faster.

See also: carbon-design-system/carbon-tutorial#3178

#### Changelog

**Changed**

- In the react tutorial, step 5: Use a `.cfignore` that only sends the static site.


alisonjoseph:
Thanks!

PR #620, opened 2019-11-22
fix(overviewCards): fixes breakpoints for component overview and data viz overview cards

### LINKED ISSUE #508: fix[charts & component overview pgs]: cards not sized same at breakpoints
## fix[charts & component overview pgs]: cards not sized same at breakpoints

Both the Chart types and the Component Overview page have the same card pattern; at xlg breakpoint they are the same size but at every other smaller breakpoint they are different. Also, both pages are wrong at mobile. Cards should be two up on mobile and bleed full screen.

@alisonjoseph Mike found this last week, not sure who to give it to... not super high priority but it does need to be fixed.

**Xlg breakpoint (same size cards); this is correct:**

image

**Large breakpoint (cards are different):**
Should be three up, Component Overview is correct here

image

**Tablet breakpoint (cards are different):**
Components overview page is correct... should be two up here and the margin is 16px (Chart types margin is wrong, it is 32px)

image

**Mobile (cards are different, both are wrong):**
Cards should be two up on mobile and bleed (like charts page but without the margin)

image

jnm2377:
Closes #508 

**Note:**
- ~added the className `bx--col-max-*` as a temporary fix for the max breakpoint in the data viz cards since there's a bug in the gatsby theme that sets the incorrect className for `colMax`.~
- ~I am unsure about how to make the data viz cards bleed in mobile since the padding is coming from the grid container. I am not entirely sure how the component overview cards do it. Any ideas @vpicone @alisonjoseph ?~


alisonjoseph:
```
.overview-card-group {
    margin: 1rem 0 2rem;
}
```

The no-gutter class is correct but this styling is breaking the bleed at mobile on the charts page. It's overriding the default row margin of -1rem left/right.

jnm2377:
@alisonjoseph thanks for the quick reply! I was confused by that when I inspected it earlier bc I saw that the side margins were `0`, but I didn't realize it's supposed to have negative margins without this override. Fixed!

vpicone:
@jnm2377 the max fix got merged upstream just waiting for a few other pr's to close before publishing the patch

jnm2377:
@vpicone gotcha. yeah, i realized the breakpoint changing at different times is just because it's not getting the media query bc of that bug. The temporary work around (of giving it the classname) helps it get the styles but it's not media query specific so not quite a true fix..

jnm2377:
@jeanservaas this is ready for re-review.

jnm2377:
@jeanservaas fixed!

PR #696, opened 2019-12-17
fix: full width editor

### LINKED ISSUE #693: Component demo: code block span full width
At the moment the code block is nested to the left of the modifier panel. After talking to Mike, we want to make the code block span the full width again and the modifier panel sit above the code block.

As is:
image

To-be: 
image

vpicone:
Closes #693 

I added a fake preview link to show the zamboni effect: https://carbondesignsystem-hbqt61mfi.now.sh/components/button/code/


vpicone:
@aagonzales oof

vpicone:
@aagonzales here's a demo with the zamboni in action: https://carbondesignsystem-hbqt61mfi.now.sh/components/button/code/

Safari should be fixed as well now

PR #707, opened 2019-12-17
docs: login pattern

jillianhowarth:
Draft of the login pattern. Presented to Governance Team, Dec 20, 2019. 

Content: Jillian and Jan
Images: Jeannie

SME feedback
- [x] Sheina Tabone
- [x] Alex Swain
- [x] Maui Francis
- [x] Luke Firth
- [ ] Peter Vachon
- [ ] Chris Connors

Scheduled for board review, December 20.


janchild:
Feedback from Luke Firth on early draft, Dec 17

**Luke Firth 4:59 PM**
Hey! I just took a look.
I think its pretty great stuff. 

Couple of notes:
The major dissonance I'm getting from it is the split between the 'simple standard single column' style, which aligns with most of the spirit and guidance included. And the Marketing split screen option. I read it like 'this is the way things should be for this reason, don't have distractions, don't have marketing... or you could use split screen with all this distracting marketing content'. Feels like perhaps the single column layout is the prefered one... but the marketing option is kind of being included because it has to be. If that makes sense - I don't see much explaining why you would use it, aside from 'if you have to', but a lot of guidance that would suggest not to use it.
I'd be curious on the research around if only including the product name changing between different IBM products is enough to clearly and quickly recognize what you are signing into when everything else is the same. My team looked into this a lot and were toying with the addition of a small app-icon above the product name to reinforce it. 

Generally, thats about it really at this point :slightly_smiling_face:

janchild:
Feedback from Alex Swain, Dec 17, 2019

**Alex Swain: 10:12 PM**

Hey Jan, some quick feedback…

I take it we are only focusing on SSO-based login for now, because Cloud Pak for Data uses a regular login experience (non-IBMid).

Feedback
* The guidance here is good, covers the majority of what would be needed for what is shown (SSO)
* We should add error states when a login fails
* We may want to reconsider the phrase of “It’s not recommended to include any distracting information or imagery to the form.” This is true but utilizing imagery or illustration/graphics in a non-SSO login can help make it nicer. Unless this is specifically guidance for SSO and other options will be added later.

janchild:
Feedback from Maui Francis, Dec 17, 2019

**Maui Francis 8:58 AM**

Feedback
* For IBMid, is there a way to show an example of the differences between what you're describing?
* I agree with Alex on the images, encouraging a particular way to inject personality is a great way to get consistency across all login experiences. e.g. the empty state blocks and how we're offering a couple of different layouts.
* I think we assume that everyone knows what these acronyms mean (SSO, URX) even terms like federated/non-federated.
* "When users enter their auth information progressively, the cognitive load is reduced...." this entire paragraph was a bit hard to follow. I had to read it 3 times to understand what it was trying to say. Is there a way to make what you're describing more simple? The first step was done great: * "On the first page, the user enters their username, then clicks Continue. The system validates the username and provides a field for the password only if required."
* Are we using "Log in" as two words? I see it refered as one word and two throughout this document.
* What is "motor load"?
* Under Best Practices, did you mean to say "...types on content on each side?"
* The last bullet point under "Do" is a series of questions that I don't quite understand. Are these places holders?
* What about the system remembering login information?
* There's guidance on the marketing content / login split, but what about if it's just personality that's being used with no marketing content? Would the login still remain on the right, or would we want our users to see the login on the left so it's the first thing they interact with as they are going left to right? I can give an example if you like.
* You have definitions of the different logins (single vs. split) but I think it's helpful if you understand when to use them, and we tell them that upfront like we've been doing with our other patterns..."When you only have a simple entry point, use a single column layout. It provides..."

janchild:
Testing results shared by Sheina Tabone:

https://ibm.ent.box.com/notes/576364139017?s=dq860ehhgum5hj4gwi9aqfe3hyxts354

connor-leech:
@jillianhowarth This looks so fantastic. Be sure to add this to the overview table 🎉

src/pages/patterns/login-pattern/index.mdx
```diff
@@ -0,0 +1,181 @@
...
+<Caption>Step 2 of the login flow for non-federated users</Caption>
+
+</Column>
+</Row>
+
+## Design types
+
+The Login page should focus on the main task of accessing one’s account. The point of this page is to get the user into their account quickly without complicating the workflow. It’s not recommended to include anything that will create too much of a distraction from the form.
+
+Simple illustrations can add personality and help differentiate the form from other products, but they should not pull the user's focus away from logging in. Consider the simplest approach to logging into an account on mobile and maintain that simplicity on tablet and desktop.
```
designertyler: Same questions about "logging in to" vs. "loggin into"

```suggestion
Simple illustrations can add personality and help differentiate the form from other products, but they should not pull the user's focus away from logging in. Consider the simplest approach to logging in to an account on mobile and maintain that simplicity on tablet and desktop.
```

connor-leech: Looks like "log in to" is generally thought to be the better move: https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to

src/pages/patterns/login-pattern/index.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+</AnchorLinks>
+
+## Overview
+
+This pattern covers the username and password screens seen by users as they log into IBM products. Federated and non-federated authentication are included in this pattern. Error states, forgot username/password, and additional authentication methods will be covered in future iterations of this pattern.
+
+| Term                 | Definition                                                                                                  |
+| -------------------- | ----------------------------------------------------------------------------------------------------------- |
+| Federated users      | IBM employees and other enterprise users who use single sign-on (SSO) to quickly sign in.                   |
```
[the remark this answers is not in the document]
jillianhowarth: I think that makes sense. Will change in the next commit!

PR #708, opened 2019-12-17
chore(project): update to v10.9.0

joshblack:
This PR is currently introducing the Release Candidates for our v10.9 release. When v10.9 moves to stable, we'll also update this to include v10.9 latest! (Unless we want to merge this and then do a separate one)

#### Changelog

**New**

**Changed**

- Update dependencies to v10.9.x RC variants

**Removed**


vpicone:
Adding `@carbon/icons-react` rc fixed the issue. There's also no duplication of bundle now that the `carbon-components-react` icons dep is no longer pinned.
Screen Shot 2019-12-18 at 2 20 24 PM
Screen Shot 2019-12-18 at 2 20 32 PM

PR #723, opened 2019-12-19
Pattern template update

janchild:
This updates the pattern template in the *How to contribute* section and also the Markdown template.


docs/pattern-template.mdx
```diff
@@ -42,7 +42,7 @@ This is an ordered list:
 
 - [Overview](#overview)
 - [The solution](#the-solution)
-- [Related patterns and components](#related-patterns-and-components)
+- [Related](#relateds)
```
designertyler: It looks like there was an "s" left over

```suggestion
- [Related](#related)
```

PR #757, opened 2020-01-08
fix: don't use css transitions while the page is still hydrating

vpicone:
This is really a major issue only on sites with an extensive load time and even then only on mobile/tablet.


vpicone:
moving to theme: https://github.com/carbon-design-system/gatsby-theme-carbon/pull/638

PR #787, opened 2020-01-15
Update CODEOWNERS

vpicone:
This should default new pr's reviews to aux squad, unless content is being updated, in which case the systems squad will be added for review.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,11 @@
 # This file determines the code owners for various directories within the IBM Carbon website.
 
+# Default to the aux team
+* @carbon-design-system/developers-auxiliary
+
+# If updating content (mdx) opt for systems team
+*.mdx @carbon-design-system/developers-system
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,11 @@
 # This file determines the code owners for various directories within the IBM Carbon website.
 
+# Default to the aux team
+* @carbon-design-system/developers-auxiliary
+
+# If updating content (mdx) opt for systems team
+*.mdx @carbon-design-system/developers-system
```
alisonjoseph: @carbon-design-system/content do you want to be auto-tagged for reviews in all mdx updates on the website? what about @carbon-design-system/design ?

vpicone: makes sense, trying to avoid bringing in the whole circus for things like this https://github.com/carbon-design-system/carbon-website/commit/e63ff51936eab06b19308d543ed67e302f44c8a6

But content definitely makes sense.

PR #790, opened 2020-01-16
Change single quote to apostrophe

### LINKED ISSUE #741: update single quote to an “apostrophe”
https://www.carbondesignsystem.com/resources#featured-stories

Screen Shot 2020-01-04 at 6 04 15 PM

xxxle0:
Closes #741

replace single quote by apostrophe


vpicone:
@xxxle0 if you need to do something like this in the future you can just type it: `’` no need for the unicode mess.

xxxle0:
yeah, my keyboard dont have `’` so i thought we need unicode for it. I will create PR for replace it. Can you merge it @vpicone ?

PR #798, opened 2020-01-16
chore: switch to nightly deploys

vpicone:


.github/workflows/deploy.yml
```diff
@@ -1,9 +1,8 @@
 name: Deploy website to IBM Cloud
 
 on:
-  push:
-    branches:
-      - master
+  schedule:
+    - cron: '0 8 * * *'
```
vpicone: 2am central time

joshblack: [Fun site](https://crontab.guru/#0_8_*_*_*) 👀

vpicone: I used that!! Haha

PR #818, opened 2020-01-21
changed 'pictogram' to 'asset' in icon contribute.mdx

chrisconnors-ibm:
changed 'pictogram' to 'asset' in line 90


chrisconnors-ibm:
wrong starting point.

PR #845, opened 2020-01-24
chore(project): update carbon to v10.9.3

joshblack:
This PR is currently introducing the Release Candidates for our v10.9.3 release. When v10.9.3 moves to stable, we'll also update this to include v10.9.3 latest! (Unless we want to merge this and then do a separate one)

The changelog for the current release is available [here](https://github.com/carbon-design-system/carbon/releases/tag/v10.9.3)

#### Changelog

**New**

**Changed**

- Update dependencies to v10.9.3 RC variants

**Removed**


vpicone:
@joshblack If we do a release with the text input size variants being different than the dropdown that’ll be a breaking change to fix.

joshblack:
Did those make it into stable? Thought they would be under `feat` and in the next minor?

tw15egan:
PR to change them is here: https://github.com/carbon-design-system/carbon/pull/5167

Text input sizing is a `feat` anyway so it wasn't going to be in this patch.

vpicone:
@tw15egan @joshblack oh right, didn’t notice the feat tag. Thanks for fixing so quickly!

vpicone:
Looks like three icons are only being exported at 16px: https://github.com/carbon-design-system/carbon/issues/5224

vpicone:
Updated to `.3` in a seperate PR, opened carbon-design-system/carbon#5224 to resolve missing icon issues upstream

PR #852, opened 2020-01-28
Change max-width of color switcher

### LINKED ISSUE #850: Theme switcher jumps at the larges size
The theme switcher on the Color/usage page jumps off the grid when it floats on the largest break point.

Steps to recreate:
Go to:
https://carbondesignsystem.com/guidelines/color/usage/

Make sure the browser width is 1584px or larger

Scroll down and watch the theme switcher

Tabs the same width as the color token table:
image

Tabs become narrow when they float
image

xxxle0:
Closes #850 

Change value of max-width color switcher

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


xxxle0:
Because we add position: fixed, we need to calculate max-width manually. Any ideas are wellcome. Thanks.

vpicone:
@xxxle0 tweaking all those percentages feels weird, but since this component is only used on one page in one site, I think it's fine to be a bit hacky. thanks!

PR #858, opened 2020-01-28
Docs: minor notification pattern updates

jillianhowarth:
Minor updates to clarify the notification pattern based on some questions posted in Slack.


src/pages/patterns/notification-pattern/index.mdx
```diff
@@ -186,6 +194,7 @@ Banners take over the top of an interface to show notifications general to the p
 #### Best practices:
 
 - Banners should be placed at the top of the content area they relate to.
+- Do not cover other content with a banner notification.
 - Place system-wide messages directly below the main header or navigation bar.
 - Banners are not sticky and should scroll with the other content on the page.
 - Only show one banner at a time.
```
tw15egan: >Carbon currently does not have a banner component, but if your team has developed one it can be contributed and made available to others.

Since we do have a banner (styled exactly the same as inline notification), should this line be removed?

janchild:
Comments in from Alex Swain: 

Alex Swain  1:17 PM
Hey Jillian and Jan, we were building out specifics for our notification pattern and came to the conclusion that the Carbon guidance is really quite comprehensive (yay!) making it a moot point for us to supplement it.

However, we noticed a couple of things that would benefit from clarification or slight adjustments to consider regarding toasts.

The timestamp that is always shown (within toasts) is normally not needed for most instances, as a toast notifies people right away. So generally the timestamp would be a second ago or similar. From what I hear, this is actually optional but designers are seeing it as something that has to be shown, may be good to call it out as an optional element.

The guidance around only showing two lines of text is good when there is a timestamp, but when no timestamp is needed, we could utilize that extra space and say 3 lines of text can be displayed rather than 2. 

The guidance around the ability to expand toasts (show more) can be problematic, as a toast is very small to begin with and at times we have no control of what error messaging comes back (length can be v-e-r-y long) from databases or servers. Toasts are really meant for at-a-glance awareness, but mentioning that the toast can be expanded to show the full text won’t be able to always handle situations where the error is more than 4-5 lines or when there are extra details. Expectation would be to have a “View details” link that would launch a dialog with the full details that can handle things more elegantly. Or at the very least, should be an alternative option.

src/pages/patterns/notification-pattern/index.mdx
```diff
@@ -100,6 +100,8 @@ Notification status helps convey the information being communicated. These statu
 
 ![Example of notification status](/images/1_notification_status_1120.png)
 
+<Caption>Notificaiton status using the inline notification component</Caption>
```
emyarod:
```suggestion
<Caption>Notification status using the inline notification component</Caption>
```

src/pages/patterns/notification-pattern/index.mdx
jillianhowarth: Good question! We're going to clarify the language about not having the component in Carbon - the inline banner is just an example to show the idea, but we haven't done enough iteration and testing to define it as a banner component.

PR #877, opened 2020-02-03
Remove design only badge from Horizontal Bar card

theiliad:
Sorry for having to make another PR for this, just noticed that the card says `design only` which no longer applies to the horizontal bar charts 🙂 

image


vpicone:
I'm lookin into the deploy preview failure, sorry about that.

theiliad:
> I'm lookin into the deploy preview failure, sorry about that.

**Np! Don't stress too much.**

We've got the demos already showing anyways, the `design only` badge should ideally be removed soon but doesn't cause massive issues for us

PR #883, opened 2020-02-07
add table-layout to make cell equal

xxxle0:
Closes https://github.com/carbon-design-system/gatsby-theme-carbon/issues/708

I add table-layout: fixed to make every column have same width.

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/src/components/ComponentStatus/component-status.scss
+++ b/src/components/ComponentStatus/component-status.scss
@@ -2,6 +2,10 @@
   white-space: nowrap;
 }
 
+.component-status .page-table {
+  table-layout: fixed;
+}
+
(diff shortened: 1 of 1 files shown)


xxxle0:
oops, Ok let me fix it

xxxle0:
hi @vpicone, should we split this table to 5 equal column ? Right now, the column width depend on the column width content.

vpicone:
@xxxle0 it's in general I think it's probably better just to stick to what the designer was asking for, thanks again for the PR!

PR #889, opened 2020-02-10
docs: use InlineNotification's in all tutorials

mattrosno:


mattrosno:
@lee-chase @cal-smith I cleaned up all of the tutorial notes and pro tips as `InlineNotification`s. Can you review your tutorials to make sure everything still looks good?

lee-chase:
@mattrosno one thing that does look more off than before is the code snippets and images busting out of the column used for the other text content.

mattrosno:
@lee-chase I agree, code snippet hangs into the gutter and the InlineNotification does not, giving this a jarring/jagged effect. I'll open an issue in the Gatsby theme to see if this is working as intended.

image

PR #904, opened 2020-02-12
docs: glossary link rewrite

connor-leech:
Adding glossary rewrite to "Action labels page" and grouping rewrites by site category.


connor-leech:
@vpicone I wouldn't have thought so either, but alas I have already heard from someone who did 😂. I think it's also referenced on some PAL sites.

We pulled outdated glossary items that we don't support anymore and refocused the section to just address UI action labels (which most were).

PR #915, opened 2020-02-17
chore: add data-table, dropdown, and date-picker

vpicone:


src/pages/components/data-table/code.mdx
```diff
@@ -4,62 +4,222 @@ description: The data table component allows for the flexible display and sortin
...
+  scope={{ rowData, headerData, React }}
+  links={{
+    React:
+      'http://react.carbondesignsystem.com/?path=/story/contentswitcher--default',
+    Angular:
+      'https://angular.carbondesignsystem.com/?path=/story/components-content-switcher--basic',
+    Vue:
+      'http://vue.carbondesignsystem.com/?path=/story/components-cvcontentswitcher--default',
+    Vanilla: 'https://the-carbon-components.netlify.com/?nav=content-switcher',
+  }}>{`<DataTable
```
jnm2377: I was also just messing around with this on a branch in my dev env before I saw this PR. I added these knobs to the default data table component demo:
```
  knobs={{
    DataTable: ['isSortable'],
    Table: ['size', 'useZebraStyles'],
  }}
```
The `isSortable` knob is working perfectly. However, I realized the Table knobs are being injected at the wrong place. Likely a bug in the regex. It's inserting the props in the middle of  `TableContainer` since it's the first "Table" string, so then it ends up looking like: 

```
Table useZebraStylesContainer
```
which doesn't work obviously. I can work on updating the regex and adding these knobs after this PR is merged if you want, or you can add it in this PR. Just thought it would be a good idea to have those props.

vpicone: Ah, yeah I gotta update the storybook links too.

jnm2377:
The data table knobs are working correctly now 🎉 👍  the selection is still not working however. It looks like the row isn't getting the class it needs when it's selected. I wonder if this is because it's added via prop... but I can't figure out where that prop is updated...

``` js
  const className = cx(props.className, {
    [`${prefix}--data-table--selected`]: props.isSelected,
  });
```
https://github.com/carbon-design-system/carbon/blob/e78bbb5cc6321c54eecd67af1d26c3bef77ad68e/packages/react/src/components/DataTable/TableRow.js#L20

**expected:**

Screen Shot 2020-02-18 at 10 16 10 AM


**actual:**

Screen Shot 2020-02-18 at 10 16 49 AM

vpicone:
@jnm2377 looks like I was missing a prop-getter for the row, fixed now.

PR #941, opened 2020-02-24
chore(tutorial): update step-1 to use ESM for carbon

joshblack:
Based on: https://github.com/carbon-design-system/carbon/issues/4847#issuecomment-590198880

This PR updates our tutorial to use ESM imports instead of `lib`. This helps with tree-shaking icons that are used by `carbon-components-react`.

#### Changelog

**New**

**Changed**

- Update step-1 to use `es` for `carbon-components-react`

**Removed**

### DIFF
--- a/src/pages/tutorial/react/step-1.mdx
+++ b/src/pages/tutorial/react/step-1.mdx
@@ -256,7 +256,7 @@ import {
   HeaderGlobalBar,
   HeaderGlobalAction,
   SkipToContent,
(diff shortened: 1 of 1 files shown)


joshblack:
Question for reviewers, would I need to update these snippets in subsequent steps and on the tutorial repo itself?

aledavila:
Yes on the tutorial repo as that is what the people are pulling down and working off of our steps we wrote and not their own work. I don't think subsequent steps show past code so I think those are fine.

PR #950, opened 2020-02-25
chore: use blob instead of copy strat

vpicone:
Uses blob strategy (like IDL) to generate SVG markup that is identical to the equivalent React markup.

Testing: SVG downloads should still work


sstrubberg:
I can check this by ensuring that the icons can still be searched for?

vpicone:
@sstrubberg this has to do with their downloading.

sstrubberg:
Download works like champ. However, there is an odd, what feels like half-a-second delay when I hover over the download icon.

Feb-27-2020 10-18-51

vpicone:
@sstrubberg hmmm, i'm not getting that. What browser?

Issue #961, opened 2020-02-27
docs(component status): Component status page is stale and updates are not automatic


connor-leech:
## Overview

The [Component status page](https://www.carbondesignsystem.com/components/component-status) has not been updated to reflect the stated version at the top of the page. At the moment, this page requires manual updates and that step hasn't effectively made it into our development workflow.

The existing format also doesn't surface status for component variations.

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/44656921/75490325-99888f80-5979-11ea-9038-dfa5c12e142f.png">

### Research

@jeanservaas has already done extensive exploration into redesigning this page. We need to determine whether the data can be populated and updated automatically via GitHub's API.

Other design systems have come up with novel solutions to the same issue and we can look to them for inspiration.

### Immediate solution

We need a near-immediate solution to this page. An inline notification along with a CTA resource link to this issue would probably solve our problem in the near term while we build out a more robust solution.

connor-leech:
## Polaris

[Link](https://polaris.shopify.com/components/component-status#navigation)

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/44656921/75490767-672b6200-597a-11ea-940a-a408dfb407f3.png">

connor-leech:
## Thumbprint

[Link](https://thumbprint.design/components/overview/)

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/44656921/75490868-8fb35c00-597a-11ea-8cc2-8ccd8fc2b708.png">

connor-leech:
## Quickbooks

[Link](https://designsystem.quickbooks.com/bolt/component-visual-dashboard/)

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/44656921/75490967-bec9cd80-597a-11ea-8f2c-8f2a2c9101d0.png">

connor-leech:
## UK.GOV

[Link](https://design-system.service.gov.uk/community/backlog/)

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/44656921/75491038-e02ab980-597a-11ea-8986-2354bb1bf4ab.png">

PR #964, opened 2020-02-28
docs: Added note and fathom tracking to component status page

connor-leech:
Added an accuracy warning and fathom tracking link to component status page. Related to: https://github.com/carbon-design-system/carbon-website/issues/961


connor-leech:
@emyarod I don't think we can test that, not the end of the world if the tracking link doesn't work here.

PR #970, opened 2020-03-02
fix: update svg packages, migrate metadata, lazy load svg components

vpicone:


sstrubberg:
It works, but the animations around hover effects on Chrome are glitchy

PR #982, opened 2020-03-05
docs(pictograms): add links to packages

alisonjoseph:
Ref #963 

Adds resource cards linking to the pictograms packages similar to Icons


alisonjoseph:
@vpicone I was just matching how we reference them everywhere else on the website. I have a meeting w/ Jan/connor to talk in more detail about elements docs on the website next week.

alisonjoseph:
Ohh I see what you're saying, they aren't in elements. Good catch

PR #993, opened 2020-03-06
feat: update component demos

jnm2377:
Updates the rest of the component demos, minus UI shell.


jnm2377:
> Seems like when there are a lot of modifiers in the right column we can't scroll down far enough to select the last item, might need some extra padding at the bottom?. (you can see this on the form page)

Added bottom padding for this. 

> Ordered list - I know the prop is nested, but this seems weird/confusing to show without it actually being nested inside another list.

I decided just to take out the knobs and show a nested list to begin with, since the regex is always going to apply the prop to the first list it finds which won't be the one that needs to be nested. 

> Modal - shows up behind left nav

Added z-index overriding styles for this. 

> Structured list - modifiers don't seem to do anything visually?

Seems like the border prop doesn't actually do anything in the component itself. Might have been something that worked in v9 but there are no selectors for it in v10 styles anymore, even though we still have ` [${prefix}--structured-list--border]: border,` in our React code. I've removed the knob and will also open an issue for it in the monorepo.

I also removed the label knob since it doesn't have a visual change. It changes the DOM element, it'll be a `div` by default for the row, but this changes it to `label`. 

The `noWrap` prop wasn't working bc the regex for the knobs works on the first item it finds, and the first row cell didn't have long enough text to see the difference (it's being applied to the first table cell in the header), so I've also removed that for now.

alisonjoseph:
Everything looks good except I still can't select the last item on the form page. 

Screen Shot 2020-03-11 at 2 45 37 PM

PR #996, opened 2020-03-09
chore(dialog): update language in dialog pattern

joshblack:
Opened this up as a discussion around switching from "modeless" to "non-modal" in our dialog pattern.

It seems like there may be conflicting sources of truth so might be helpful to sync up and see which one makes most sense moving forward. Most likely the best sources we have for "non-modal" dialog include:

- [WAI ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal)
  - They mention non-modal dialogs in their guidance but don't explicitly talk about them
- [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role)
  - "Dialogs can be either non-modal ... or modal"
- [Nielsen](https://www.nngroup.com/articles/modal-nonmodal-dialog/)
  - Uses `nonmodal` as the primary terminology but does mention `modeless`

Links that use modeless
- [Apple](https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/dialogs/)
  - They use modeless to apply to windows in macOS
- [Microsoft](https://docs.microsoft.com/en-us/windows/win32/dlgbox/about-dialog-boxes#modeless-dialog-boxes)
  - They use modeless to apply to windows in Windows

It seems from some quick research that modeless may be prevalent on desktop but non-modal is the preference on the Web. Could easily be wrong here, though!

#### Changelog

**New**

**Changed**

- Update `modeless` to `non-modal`

**Removed**

### DIFF
--- a/src/pages/patterns/dialog-pattern/index.mdx
+++ b/src/pages/patterns/dialog-pattern/index.mdx
@@ -5,7 +5,7 @@ description: A dialog is a “conversation” between the system and the user. I
 
 <PageDescription>
 
-A dialog is a “conversation” between the system and the user. It is prompted when the sytem needs input from the user or to give the user urgent information concerning their current workflow. There are two types of dialogs, modal and modeless.
+A dialog is a “conversation” between the system and the user. It is prompted when the sytem needs input from the user or to give the user urgent information concerning their current workflow. There are two types of dialogs, modal and non-modal.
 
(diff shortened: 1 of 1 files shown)


connor-leech:
This makes sense 👍

PR #998, opened 2020-03-09
docs: Update pictogram link

connor-leech:
Contribution link pointed users to defunct repo. Link now directs users to the correct location.

P.S. This copy needs some love and we should standardize the guidance between the IDL site and the Carbon site.

@joshblack This contribution process is quite similar to the icons contribution process, correct?


connor-leech:
Related issue: https://github.com/carbon-design-system/carbon-website/issues/999

joshblack:
@connor-leech yup! Should be almost identical if I remember correctly.

PR #1009, opened 2020-03-12
fix(docs): field-02 usage

### LINKED ISSUE #575: Role of `$field-02` token
`$field-02` role column reads “Fields on $ui-02 backgrounds” . This yields a field the same color as it’s  background.  Seems as if it should read “Fields on $ui-01 backgrounds” instead.

## Steps to reproduce the issue
Visit the Carbon Design System [color usage page](https://www.carbondesignsystem.com/guidelines/color/usage/)

image

Qs? contact @Maezlin-Avila-Taylor

aagonzales:
Closes #575


src/data/guidelines/color-tokens.js
```diff
@@ -502,7 +502,7 @@ const colorTokens = {
       },
     },
     '$field-02': {
-      role: ['`Light` variant input fields', 'Fields on $ui-02 backgrounds'],
+      role: ['"Light" variant input fields', 'Fields on $ui-01 backgrounds'],
```
vpicone: Need to manually do smart quotes when it's not an `mdx` file.

```suggestion
      role: ['“Light” variant input fields', 'Fields on $ui-01 backgrounds'],
```

PR #1014, opened 2020-03-12
docs(overflow-menu): adds a11y overflow menu docs

aledavila:
Reference #491 

Adds a11y overflow menu docs

Make sure content matches https://pages.github.ibm.com/IBMa/Carbon_Integration/Doc/OverflowMenu-Doc.html


src/pages/components/overflow-menu/accessibility.mdx
```diff
@@ -0,0 +1,222 @@
...
+</AnchorLinks>
+
+## How it works
+
+The Carbon overflow menu component is used when there is limited space to provide a pop-up list of actions that are available to the user. The overflow menu button uses `tabindex="0"` to ensure it is in the tab order for keyboard only users. When the button has focus, Space key as well as the Enter key activates the pop-up menu. WAI-ARIA roles and states are used to ensure the component is accessible to people using assistive technologies. The pop-up menu has the ARIA role of `role="menu"`, a label of `aria-label="Menu"` and includes `aria-haspopup="true"` to let the user know it is a pop-up menu. When the menu button is collapsed the pop-up menu is hidden so the ARIA state is set to `aria-expanded="false"` to advise users of assistive technologies that the pop-up menu is not visible. When the pop-up menu is expanded the ARIA state changes to `aria-expanded="true"` and the menu is displayed. Each menu option contains an ARIA role of `role="menuitem"` and unique ARIA labels for each option. When focus is on each option there is a prominent style change to the border to ensure the focus is clearly displayed.
```
connor-leech:
```suggestion
The Carbon overflow menu component is used when there is limited space to provide a pop-up list of actions that are available to the user. The overflow menu button uses `tabindex="0"` to ensure it is in the tab order for keyboard only users. 

When the button has focus, Space key as well as the Enter key activates the pop-up menu. WAI-ARIA roles and states are used to ensure the component is accessible to people using assistive technologies. The pop-up menu has the ARIA role of `role="menu"`, a label of `aria-label="Menu"` and includes `aria-haspopup="true"` to let the user know it is a pop-up menu. 

When the menu button is collapsed the pop-up menu is hidden so the ARIA state is set to `aria-expanded="false"` to advise users of assistive technologies that the pop-up menu is not visible. When the pop-up menu is expanded the ARIA state changes to `aria-expanded="true"` and the menu is displayed. 

Each menu option contains an ARIA role of `role="menuitem"` and unique ARIA labels for each option. When focus is on each option there is a prominent style change to the border to ensure the focus is clearly displayed.
```

connor-leech:
This looks great, just added some line breaks in the first paragraph!

PR #1046, opened 2020-03-25
Removed the component table, added explanation

shinytoyrobots:
Relates to #961.

Given that the component status table is inaccurate, we should take it down entirely. Then expand on the additional notification added by #964 to provide extra information.

NB: Feedback on the last sentence of my text;

> All of the components you see published under the ‘Components’ section on the Carbon Design System site are production ready and approved for use in React.

I think that's OK, but does it imply non-approval for other frameworks?


janchild:
Reviewed and updated the text directly. Should be good to go, @shinytoyrobots !

PR #1052, opened 2020-03-26
docs: Community notification

connor-leech:
DO NOT MERGE

In light of the recent community updates, I've added a community tag to all pages maintained by community teams.

Question: Would it be worth adding some sort of indicator for the tiles on the Resources page?


janchild:
@connor-leech I don't think we need the third line here. The explanation in the first sentence and the link to contact the maintainers does the trick!

Issue #76, opened 2019-08-15
Toast notification positioning guidance


designertyler:
Add additional guidance and visual that explains how the toast notification should be positioned in relation to the grid and header.

Also, include direction on when to show toast in the upper right vs. bottom right (when it could cover nav or other status when it's close to the affordance to bring back the notification)

![image](https://user-images.githubusercontent.com/19270502/59061280-b66bb880-8868-11e9-870f-d021c9b6796f.png)

aagonzales:
@jillianhowarth this could be an item to address with the notification usage updates you are doing.

aagonzales:
This has been addressed at a high level with the reset update but could be clarified with specific measurements on the style tab page.

PR #1081, opened 2020-04-02
Notification usage updates

jillianhowarth:
Updates the notification component usage guidance to follow the new template.


src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,166 @@ description: Notifications are messages that communicate information to the user
...
+![High and low contrast inline notifications](images/notification-usage-4.png)
 
 </Column>
 </Row>
+
+### Action
+
+**Inline notification**
+
+Inline notifications have an optional ghost button action that is right aligned. This button should allow users to take further action on the notification.
```
joshblack, line 154: > Inline notifications have an optional ghost button action that is **right aligned**

What would be the best way to talk about direction if we do left-to-right and right-to-left? Or is this is saying that the action is always on the right-hand side?

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,166 @@ description: Notifications are messages that communicate information to the user
...
+
+**Toast notifications**
+
+Toast notifications can include a link at the end of their body content. This link should be short and navigate users to a page or modal where they can take action to address the notification or find further information. Because toast notifications automatically dismiss, it is important that there are alternative routes to navigate to the link destination.
```
joshblack, line 158: > Because toast notifications automatically dismiss, **it is important** that there are alternative routes to navigate to the link destination.

Should we potentially include stronger language around this in the sense that they must provide an alternate way to view the notification?

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,166 @@ description: Notifications are messages that communicate information to the user
...
-Inline notifications appear near its related item. In [forms](/components/form/code), we recommend placing the inline notification at the bottom of the form, right before the submission buttons. Depending on the context of the page, inline notifications can appear above the content as well.
+### Body content
+
+- The body of the notification should be no more than two lines.
+- Be concise and avoid repeating or paraphrasing the title.
+- Explain how to resolve the issue by including any troubleshooting actions or next steps. You can include links within the notification body that redirect the user to next steps.
+
+### Action
+
+- Keep labels consice and clearly indicate the action the user can take.
```
joshblack, line 125:
```suggestion
- Keep labels concise and clearly indicate the action the user can take.
```

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,166 @@ description: Notifications are messages that communicate information to the user
...
+### Title
+
+- The title should be short and descriptive, explaining the most important piece of information.
+- When possible, communicate the main message using just the title.
+- Use a period only if the title is a full sentence.
 
-Inline notifications appear near its related item. In [forms](/components/form/code), we recommend placing the inline notification at the bottom of the form, right before the submission buttons. Depending on the context of the page, inline notifications can appear above the content as well.
+### Body content
+
+- The body of the notification should be no more than two lines.
```
joshblack, line 119: How are we suggesting folks tackle this in the globalization/overflow use-case?

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,166 @@ description: Notifications are messages that communicate information to the user
...
+- Be concise and avoid repeating or paraphrasing the title.
+- Explain how to resolve the issue by including any troubleshooting actions or next steps. You can include links within the notification body that redirect the user to next steps.
```
joshblack, line 121: How are we anticipating keyboard or screen reader users to navigate to the toast if there is interactive content inside? By default, it seems like these might be hard to reach quickly within a small time window.

src/pages/components/notification/usage.mdx
aagonzales: We dont really address this topic anywhere and at the moment we dont really have a strategy for it. This is something we should talk about though.

aagonzales: Speaking of the action button placement though this week we are adding a small width behavior change to it. So it sometime also on the bottom.
https://github.com/carbon-design-system/carbon/pull/5763

src/pages/components/notification/usage.mdx
aagonzales: Yeah good point with translation we can't really control the line length. We could make these comments less line specific. Like "Content should be 1-2 short sentences." I do think we should emphasize somehow that it shouldn't have excess content. I could even say one short sentence to emphasize that.

src/pages/components/notification/usage.mdx
aagonzales: I think this is more talking to inline notification that toast. But i think it would be good to emphasis that point here. Like 

> "If you include links within the notification body that redirect the user to steps, be sure keyboard and screenreader user have enough time to access the notification or can easily find the notification after it disappears."

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
+- Use a period only if the title is a full sentence.
+
+#### Body content
 
-| Notification type     | Purpose                                                                                                                                                               |
-| --------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| _Toast notification_  | Toasts are a non-modal, time-based window elements used to display short messages; they usually appear at the bottom of the screen and disappear after a few seconds. |
-| _Inline notification_ | Inline notifications show up in task flows, to notify users of the status of an action. They usually appear at the top of the primary content area.                   |
+- Be concise and avoid repeating or paraphrasing the title.
+- Content should be 1-2 short sentences.
```
janchild, line 74:
```suggestion
- Limit content to one or two short sentences.
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
+- Explain how to resolve the issue by including any troubleshooting actions or next steps. You can include links within the notification body that redirect the user to next steps.
 
-### High vs. low contrast
+#### Action
 
-Use high-contrast style notifications for critical messaging. Low-contrast notifications are best used for supplemental messaging and are less visually disruptive to users. When in doubt, use the low contrast style.
+- Keep labels concise and clearly indicate the action the user can take.
+- Limit action labels to two words or less.
```
janchild, line 80:
```suggestion
- Limit action labels to one or two words. For a list of approved action labels, see Carbon's [Content guidelines](https://www.carbondesignsystem.com/guidelines/content/action-labels). 
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
-## Placement
+### Behavior
 
-#### Toast notifications
+**Dismissal**
 
-Toast notifications slide in and out a page from the top-right corner. Actionable notifications do not appear on mobile screen widths.
+Inline notifications do not dismiss automatically. They persist on the page until the user dismisses them or takes action that resolves the notification.
+
+A small "x" in the top right corner is used to dismiss inline notifications. Including the close button is optional and should not be included in a notification if it is critical a user reads or interacts with it.
```
janchild, line 122:
```suggestion
A small "x" in the top right corner is used to dismiss inline notifications. Including the close button is optional and should not be included if it is critical for a user to read or interact with the notification.
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
 
-![Form example with inline notification](images/notification-usage-3.png)
+![High and low contrast inline notifications](images/notification-usage-4.png)
 
 </Column>
 </Row>
+
+## Related
+
+- Modals
```
janchild, line 180:
```suggestion
- [Modal component](https://www.carbondesignsystem.com/components/modal/usage/)
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
+- Notification pattern
```
janchild, line 181:
```suggestion
- [Notifications pattern](https://www.carbondesignsystem.com/patterns/notification-pattern)
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
+
+### Formatting
+
+**Action**
+
+Toast notifications can include a link at the end of their body content. This link should be short and navigate users to a page or modal where they can take action to address the notification or find further information. Because toast notifications automatically dismiss, it is important that there are alternative routes to navigate to the link destination.
+
+If you include links within the notification body that redirect the user to steps, be sure keyboard and screenreader users have enough time to access the notification or can easily find the notification after it disappears.
+
+**Timestamp**
```
janchild, line 134:
```suggestion
**Time stamp**
```

src/pages/components/notification/usage.mdx
```diff
@@ -1,82 +1,183 @@
...
+
+Toast notifications include a timestamp at the bottom the container. The timestamp shows the time the notification was sent. Using timestamps is optional but toasts should be consistent across the product so either all toasts should include timestamps or none of them should. The timestamp is optional and can be removed if a third line of content is needed.
```
janchild, line 136:
```suggestion
Toast notifications include a time stamp at the bottom the container. The time stamp shows the time the notification was sent. Using time stamps is optional but toasts should be consistent across the product so either all toasts should include time stamps or none of them should. The time stamp is optional and can be removed if a third line of content is needed.
```

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,198 @@ description: Notifications are messages that communicate information to the user
...
+- Be concise and avoid repeating or paraphrasing the title.
+- Limit content to one or two short sentences.
+- Explain how to resolve the issue by including any troubleshooting actions or next steps. You can include links within the notification body that redirect the user to next steps.
 
-#### Dismissal
+#### Action
 
-We recommend that toast notifications automatically disappear after five seconds. Inline notifications are persistent until the user dismisses them. All notifications have at least one method of dismissal (typically, it is a small “x” in the upper right hand corner).
+- Keep labels concise and clearly indicate the action the user can take.
+- Limit action labels to one or two words. For a list of approved action labels, see Carbon's [content guidelines](https://www.carbondesignsystem.com/guidelines/content/action-labels).
```
joshblack, line 80: Was surprised to see approved here, is there a process through which teams get labels approved? Or are these recommendations by our team?

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,198 @@ description: Notifications are messages that communicate information to the user
...
 
 </Column>
 </Row>
 
-#### Inline notifications
+#### Action
+
+Toast notifications can include a link at the end of their body content. This link should be short and navigate users to a page or modal where they can take action to address the notification or find further information. Because toast notifications automatically dismiss, there should be alternative routes to navigate to the link destination.
+
+If you include links within the notification body that redirect the user to steps, be sure keyboard and screenreader users have enough time to access the notification or can easily find the notification after it disappears.
```
joshblack, line 150: In general, it seems like we would want to recommend that notifications don't have any interactive content within them. When using a live region, the text of the area is what is communicated to the end user:

demo

As a result, there is no way for them to know if the content is interactive or not nor is there an easy way for them to navigate to it.

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,198 @@ description: Notifications are messages that communicate information to the user
...
+Toast notifications include a time stamp at the bottom the container. The time stamp shows the time the notification was sent. Using time stamps is optional but toasts should be consistent across the product so either all toasts should include time stamps or none of them should. The time stamp is optional and can be removed if a third line of content is needed.
+
+#### Sizing
 
-Inline notifications appear near its related item. In [forms](/components/form/code), we recommend placing the inline notification at the bottom of the form, right before the submission buttons. Depending on the context of the page, inline notifications can appear above the content as well.
+Toast notifications use a fixed width and their height depends on the length of the notification message. As noted in the content guidelines, limit toast notifications to two lines of text.
+
+#### Placement
+
+Toast notifications slide in and out from the top right of the screen. They align to the edge of the last column and can stack with `$spacing-03` in-between. New toast notifications should appear at the top of the list, with older notifications being pushed down until they are dismissed.
```
joshblack, line 162: What happens if the height of the stacked notifications is greater than the browser height?

Notifications are typically absolutely positioned, as a result, how would one implement this guidance:

> They align to the edge of the last column

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,198 @@ description: Notifications are messages that communicate information to the user
...
+
+Actionable toast notifications do not appear on mobile screen widths.
```
joshblack, line 164: What was the reasoning behind this constraint? 👀

src/pages/components/notification/usage.mdx
```diff
@@ -4,79 +4,198 @@ description: Notifications are messages that communicate information to the user
...
+![Example of toast notification placement](images/notification-placement-toast.png)
 
 </Column>
 </Row>
+
+### Behavior
+
+#### Dismissal
+
+Toasts dismiss automatically after five seconds on the screen. They can also include a close button so users can dismiss them sooner. Toasts cover content on the screen so they should always be easily dismissed. Because toast notifications dismiss automatically, users should be able to access them elsewhere after the toast disappears. This allows them to be accessible for users who need more time reading the notification or who may want to refer back to the notification.
```
joshblack, line 178: > Because toast notifications dismiss automatically, users should be able to access them elsewhere after the toast disappears

Should we word this more strongly to say that there must be an alternate location if the toast is automatically dismissed?

src/pages/components/notification/usage.mdx
janchild: These labels have been developed through multiple rounds of review with Cloud Data & AI, IoT, and Watson designers and content designers.

src/pages/components/notification/usage.mdx
shinytoyrobots: It definitely seems like it should be a requirement rather than a preference, right? (so "users _must_ be able to access..." rather than "users _should_...")

src/pages/components/notification/usage.mdx
shinytoyrobots: More specifically. Is the feedback here that this guidance should change because we don't/can't include links in notifications?

aagonzales: This is technically speaking something that we currently allow and show in the storybook. She is only commenting on what we're currently allowing. If we don't want to allow this then we need to also remove this from the component code. 

image

src/pages/components/notification/usage.mdx
shinytoyrobots: Suggestion - should read; "For a list of recommended labels...". I'm not sure "approved" is correct here, in that it implies anything else is non-approved (but we don't have any real control over the action labels different teams might use).

Process for choosing which labels are recommended is outside the scope of this PR, but we can discuss if there is something we want to document/clarify.

src/pages/components/notification/usage.mdx
shinytoyrobots: Clarification on @joshblack's last point; "absolute positioning" means that an element doesn't reference other elements, so the "align to the edge of the last column" is difficult/impossible to apply from a code perspective.

Do we need to either address the design guidance, or address the development implementation?

aagonzales: This is a good point that we don't currently have a solution for. This will need to be opened as a issue to figure out. I also think if they're disappearing after x seconds the chances of there being a long stack of them seems very unlikely.

src/pages/components/notification/usage.mdx
aagonzales: Would simply changing "should be" to "must be" be enough?

src/pages/components/notification/usage.mdx
aagonzales: I don't think it's implying that labels have to be approved before using, but merely saying the action labels on that content page are approved to use. Maybe it should say "For a list of _suggested_ action labels, ..."

src/pages/components/notification/usage.mdx
jillianhowarth: This is existing guidance we have right now. I'm not sure the original reasoning behind it but I'm trying not to changing existing guidance without having a pretty good reason to change it.

src/pages/components/notification/usage.mdx
jillianhowarth: @joshblack do you have a suggestion for how to provide the positioning guidance? I was trying to address this issue https://github.com/carbon-design-system/carbon-website/issues/76

src/pages/components/notification/usage.mdx
janchild: It needs to be stronger than just suggested. If people have concerns with "approved" then I'd like to go with "recommended".

src/pages/components/notification/usage.mdx
janchild: Yes, I think so.

src/pages/components/notification/usage.mdx
joshblack: @aagonzales definitely agreed, I think this needs to be a conversation for us all to better understand what is/is not applicable here. It feels odd given that so many sites use this pattern that this is how a screen reader behaves, maybe I'm misunderstanding something? Was surprised when I was testing it out in that recording above that it announced the way that it did.

@dakahn do you have any insights here? Clearly this is a common pattern, including notifications at the OS-level, is there anything we can do to accommodate interactive content either through a link/action button/close button?

src/pages/components/notification/usage.mdx
aagonzales: oops sorry y'all my internet is terrible and for some reason no one else comments loaded so I wasn't purposefully ignoring any earlier comments.

src/pages/components/notification/usage.mdx
joshblack: @jillianhowarth it seems like we could re-word it to say that the notifications are aligned vertically and use the spacer in between and not mention the edge alignment? Just a thought 👀

src/pages/components/notification/usage.mdx
joshblack: Got it! Makes sense 👍

src/pages/components/notification/usage.mdx
dakahn: **At IBM we have the requirement that any mouse interaction have a keyboard equivalent.** With notifications that popup and disappear (toasts or whatever) and have interactable child elements this rule becomes problematic. 

How do we move keyboard focus to that notification? If this were a modal dialog we'd simply steal the users focus and _move it to the dialog_ when the modal was opened and then trap focus in the dialog until an action was taken. Toasts come and go as the user works and are intended to be unobtrusive so the above described interaction pattern doesn't really work. 

At the application level there are ways to solve around this problem (having notifications collected somewhere that _is keyboard accessible_ like this:

image

but its always highly inadvisable to use operating system level interactions as guidance for building websites since -- in short -- they are very different contexts and only superficially comparable.

joshblack: @dakahn it seems like the guidance from WAI-ARIA gets distilled into two categories: alerts and alert dialogs:

> An alert is an element that displays a brief, important message in a way that attracts the user's attention without interrupting the user's task
> Because alerts are intended to provide important and potentially time-sensitive information without interfering with the user's ability to continue working, it is crucial they do not affect keyboard focus.

> The alert dialog is designed for situations where interrupting work flow is necessary.

---

Alerts are what we can use for time-based notifications, but if there is anything interactive then we need to use the alert dialog?

What happens if we want to use a timed alert, but also want to give someone the ability to dismiss it early? This seems like it'd directly contradict the alert guidance as it says that it shouldn't steal focus from the user.

**Links & resources**

- [Alert example](https://www.w3.org/TR/wai-aria-practices/examples/alert/alert.html)
- [Alert dialog example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html)

vpicone: Most common example I can think of is the service work content refresh confirmation or a cookie approval form.

A notification tray isn’t enough for these types of product-level notifications/approvals. A modal on the other hand is far too invasive. I don’t see the problem with adding actions to our Toast notifications so long as we prevent their automatic dismissal.

dakahn: > What happens if we want to use a timed alert, but also want to give someone the ability to dismiss it early?

They're highly problematic for sure. You can minimize a lot of these gnarly UX frustrations for keyboard/screen reader/magnification users, but some of those solutions cause frustrations for other (dialogs with interactable elements becoming modal for example solves one problem but creates another).

joshblack:
@jillianhowarth definitely didn't mean to suggest that we should solve all of that in this PR, sorry if it came across that way!

Finally caught up on that one thread in Slack and left a comment, let me know if you want to chat there or if review comments here would be preferred 👍

PR #1085, opened 2020-04-02
Docs: Notification pattern update

jillianhowarth:
Update notification status image and table to include the informational notification with an icon.


laurenmrice:
For the `Banner` section, That info notification needs the icon added too

laurenmrice:
That notification icon looks good, but now the width is too short. The notification needs to go full width since its a banner notification.  Should be the same length as the image above^

PR #1114, opened 2020-04-17
fix(accordion): usage update

aagonzales:
Accordion usage content update.

Closes https://github.com/carbon-design-system/carbon-website/issues/1064


src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,180 @@ description: The accordion component delivers large amounts of content in a smal
...
+
+<Row>
+<Column colLg={8}>
+
+![accordion placement](images/accordion-placement-2.png)
+
+</Column>
+</Row>
+
+#### Grid placement
```
joshblack, line 91: I think we still need to validate the grid placement stuff after our conversation last week. Is there a specific issue/list I can use to track which components we'd like to have work with the grid? 👀

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,180 @@ description: The accordion component delivers large amounts of content in a smal
...
+
+<Row>
+<Column colLg={8}>
+
+![accordion interactives states](images/accordion-state-3.png)
+
+</Column>
+</Row>
+
+### Click target
```
joshblack, line 139: Would we want to talk through keyboard interaction here too? (same thing just with space/enter but with focus outline instead of click target area)

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,180 @@ description: The accordion component delivers large amounts of content in a smal
...
+![accordion click targets](images/accordion-click-target.png)
+
+</Column>
+</Row>
+
+## Content
+
+### Main elements
+
+#### Title
```
joshblack, line 155: It seems that WAI ARIA recommends that the titles be treated as a specific heading level (1-6), would we want to cover that here or would it be better in a different area? 👀 

Reference: https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties

> Each accordion header button is wrapped in an element with role heading that has a value set for aria-level that is appropriate for the information architecture of the page.

aagonzales: Would this just be part of the accessibility tab content?

aagonzales: Seems more about implementation than ux usage. So I think the code tab or accessibility tab might be the better fit.

aagonzales: Now that I'm thinking of it. How does a user set a heading value of type inside of a component?

src/pages/components/accordion/usage.mdx
aagonzales: @joshblack Is there an easy way to find out what the existing keyboard commands are for a component?

src/pages/components/accordion/usage.mdx
aagonzales: I thought we agreed to leave the content in because it demonstrates the correct and indented grid placement and we'd work out how to correctly document a hack on the code tab.  

Here is the issue for grid hanging components: https://github.com/carbon-design-system/carbon/issues/5892

joshblack: @aagonzales it seemed like the process we wanted to hit was one where design and development were coordinating around proposed functionality. It seems like we're at the phase of design proposing this functionality and I thought the idea was that dev would take that issue and figure out how feasible it would be before we would offer it as guidance. The question would be if we offer the guidance before we validate it then how do we handle what happens if we end up not being able to do it.

Thanks so much for the issue, will try and see if we can get this into our April work or at the very least make sure it's in May 🙏

aagonzales: Yes, but I thought we agreed to document the design intent now as it can be achieved through other means outside of our grid package. 

That or we need change the component code, to have the rules and type alignment because we never want the accordion content to be indented from the other page content.

joshblack: @aagonzales I think we'd want to be careful around folks tackling this independently as it would require overriding styles on our components. This can cause unintended breakage down the line either due to unrelated updates for that component or updates to guidance. It'd be similar to hard coding a token instead of using a token.

src/pages/components/accordion/usage.mdx
aagonzales: Ok I added in a blurb about this.

src/pages/components/accordion/usage.mdx
aagonzales: Ok I added some content in, let know me if it looks correct.

aagonzales:
@jeanservaas Ok I updated the images. I choose the 3 across image placement. We do advise against putting an accordion in a modal in the dialog pattern docs so I don't really want to add that example. I think the 4 with the empty one could also work but seeing the in row seemed nice.

src/pages/components/accordion/style.mdx
```diff
@@ -13,6 +13,16 @@ tabs: ["Code", "Usage", "Style", "Accessibility"]
...
 
+<Row>
+<Column colLg={12}>
+
+![accordion interactives states](images/accordion-style-2.png)
+
+</Column>
+</Row>
+
+<Caption>Enabled, hover, focus, disabled</Caption>
```
janchild, line 24:
```suggestion
<Caption>Examples of enabled, hover, focus, and disabled states</Caption>
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+<AnchorLink>References</AnchorLink>
+<AnchorLink>Feedback</AnchorLink>
+
+</AnchorLinks>
+
+## Overview
+
+The accordion component delivers large amounts of content in a small space through progressive disclosure. The header title give the user a high level overview of the content allowing the user to decide which sections to read.
+
+Accordions can make information processing and discovering more effective. However, it does hide content from users and its important to account for a user not noticing or reading all of the included content. If a user is likely to read all of the content then don't use an accordion as it adds the burden of an extra click, instead use a full scrolling page with normal headers.
```
janchild, line 29:
```suggestion
Accordions can make information processing and discovering more effective. However, it does hide content from users and it's important to account for a user not noticing or reading all of the included content. If a user is likely to read all of the content then don't use an accordion as it adds the burden of an extra click; instead use a full scrolling page with normal headers.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+<Row>
+<Column colLg={8}>
+
+![accordion anatomy](images/accordion-anatomy.png)
+
+</Column>
+</Row>
+
+1. **Header**: contains the section title and is control for revealing the panel.
+2. **Icon**: indicates if the panel is open or closed
```
janchild, line 50:
```suggestion
2. **Icon**: indicates if the panel is open or closed.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+3. **Panel**: the section of content associated with an accordion header.
+
+### Alignment
+
+By default the chevron icon is placed on the `end` side of the header. This allows for the title on the `start` side to align with other type elements in the layout.
+
+<DoDontRow>
+<DoDont
+colMd={4} colLg={4}
+caption="Do: use the default icon alignment to keep accordion text aligned with other content on the page.">
```
janchild, line 60:
```suggestion
caption="Do use the default icon alignment to keep accordion text aligned with other content on the page.">
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+
+![Do: use the default icon alignment to keep accordion text aligned with other content on the page.](./images/accordion-alignment-do.png)
+
+</DoDont>
+<DoDont
+colMd={4} colLg={4}
+type="dont"
+caption="Don’t: place icons on the left and create misalignment with the other text on the page.">
```
janchild, line 68:
```suggestion
caption="Don’t place icons on the left and create misalignment with the other text on the page.">
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+
+![Don’t: place icons on the left and create misalignment with the other text on the page.](./images/accordion-alignment-dont.png)
+
+</DoDont>
+</DoDontRow>
+
+#### Modifier
+
+In rare scenarios, the accordion can be modified to place the icon in front of the title for only tree-like purposes. However, most instances should use the default alignment and for any pure content or documentation purposes it is required to use the default alignment. Icon placement should be consistent throughout your experience.
```
janchild, line 77:
```suggestion
In rare scenarios, the accordion can be modified to place the icon in front of the title to function more like a tree. However, most instances should use the default alignment and for any pure content or documentation purposes, the default alignment is required. Icon placement should be consistent throughout your experience.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+<Column colLg={12}>
+
+![accordion placement](images/accordion-placement-2.png)
+
+</Column>
+</Row>
+
+#### Grid placement
+
+When placing accordion on the 2x Grid the indented title aligns to the grid columns and the top and bottom borders hang in the gutter.
```
janchild, line 93:
```suggestion
When placing an accordion on the 2x Grid the indented title aligns to the grid columns and the top and bottom borders hang in the gutter.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+
+<Row>
+<Column colLg={8}>
+
+![multiple sections expanded](images/accordion-state-2.png)
+
+</Column>
+</Row>
+
+In addition to the collapsed and expanded states, checkboxes also have interactive states for focus, hover, disabled, and skeleton. See the [style tab](/components/accordion/style) for more details.
```
janchild, line 129:
```suggestion
In addition to the collapsed and expanded states, accordions also have interactive states for focus, hover, and disabled. See the [style tab](/components/accordion/style) for more details.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+![accordion interactives states](images/accordion-state-3.png)
+
+</Column>
+</Row>
+
+### Interactions
+
+#### Mouse
+
+Users can trigger a state change by clicking on the chevron or clicking anywhere in the entire header area.
```
janchild, line 143:
```suggestion
Users can trigger a state change by clicking on the chevron or clicking anywhere in the header area.
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+<Column colLg={8}>
+
+![accordion click targets](images/accordion-click-target.png)
+
+</Column>
+</Row>
+
+#### Keyboard
+
+Users can trigger a state change by pressing the `Enter` or `Spacebar` key while the header area has focus. For additional keyboard interactions, see the [accessiblity tab](/components/accordion/accessibility#how-it-works).
```
janchild, line 155:
```suggestion
Users can trigger a state change by pressing the `Enter` or `Spacebar` key while the header area has focus. For additional keyboard interactions, see the [accessibility tab](/components/accordion/accessibility#how-it-works).
```

src/pages/components/accordion/usage.mdx
```diff
@@ -4,20 +4,199 @@ description: The accordion component delivers large amounts of content in a smal
...
+- The title should be as brief as possible while still being clear and descriptive.
+- Each title should be wrapped in a [role heading](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties) (h1-h6) that is appropriate for the information architecture of the page.
+
+#### Body copy
+
+- Content inside of a section may be split into paragraphs and include sub-headers if needed.
+
+### Scrolling content
+
+When the accordion content is longer than the view port the whole accordion should vertically scroll. Content should not scroll inside of the an individual panel. Content should never scroll horizontally in an accordion.
```
janchild, line 184:
```suggestion
When the accordion content is longer than the viewport the whole accordion should vertically scroll. Content should not scroll inside of an individual panel. Content should never scroll horizontally in an accordion.
```

src/pages/components/accordion/usage.mdx
aagonzales: What is we talked about it from a type alignment perspective instead of grid placement. It is a fundamental design principle of the new language to have the strong vertical type alignment going down the page. What if we talked to that point instead of talking about grid gutters and hanging? I honestly don't care how it happens as long type inside the accordion and type on the page matches up.

aagonzales: We just need to advise designers that this is how content should alignment when using accordion. 

image

aagonzales: ^anything other than this alignment is unacceptable.

shinytoyrobots: That seems clear from a guidance perspective.

Unless guidance is impossible to implement, then we need to provide guidance on implementation - even (arguably, especially) - when it's unclear or hacky. The reality is, people already have to implement this guidance. We should offer a recommended or (if uncomfortable making a recommendation), a possible, implementation option.

The broader conversation of how to subsequently improve that experience (for this component, and for others) should fall outside the scope of this PR, I think.

PR #1115, opened 2020-04-17
chore: update theme and other deps

vpicone:


package.json
```diff
@@ -33,10 +33,10 @@
     "@carbon/charts": "^0.28.1",
     "@carbon/charts-react": "^0.28.1",
     "@carbon/elements": "^10.10.2",
-    "@carbon/pictograms": "^10.9.1",
```
sstrubberg, line 36: why did 10.9.1 get updated and yet... not update?

vpicone: I think that's because prettier re-ordered them to be alphabetical. Before, it came after "@carbon/elements" now it comes after "@carbon/icons-react"

vpicone:
# Heading

<a href="google.com>testing</a>

PR #1118, opened 2020-04-20
Update accessibility.mdx

joshblack:
Update accessibility docs to remove incorrect guidance around the role of an accordion.

#### Changelog

**New**

**Changed**

- Update accessibility docs to remove incorrect guidance around the role of an accordion

**Removed**


src/pages/components/accordion/accessibility.mdx
```diff
@@ -26,7 +26,7 @@ import {
...
+When an accordion panel is collapsed the content is hidden so the ARIA state, `aria-expanded="false"` advises users of assistive technologies that the accordion panel is not visible. When the accordion header is expanded the ARIA state changes to `aria-expanded="true"` and the accordion panel content is displayed. Each accordion header contains an ARIA label, and the header buttons have a aria-control property set that points to the unique id of the panel it controls. When focus is on the Accordion header there is a prominent style change to the border as well as background of the accordion header.
```
janchild, line 29:
```suggestion
When an accordion panel is collapsed the content is hidden, therefore the ARIA state `aria-expanded="false"` advises users of assistive technologies that the accordion panel is not visible. When the accordion header is expanded the ARIA state changes to `aria-expanded="true"` and the accordion panel content is displayed. Each accordion header contains an ARIA label, and the header buttons have an ARIA-control property set that points to the unique id of the panel it controls. When focus is on the Accordion header there is a prominent style change to the border as well as background of the accordion header.
```

src/pages/components/accordion/accessibility.mdx
```diff
@@ -26,7 +26,7 @@ import {
...
+When an accordion panel is collapsed the content is hidden, therefore the ARIA state `aria-expanded="false"` advises users of assistive technologies that the accordion panel is not visible. When the accordion header is expanded the ARIA state changes to `aria-expanded="true"` and the accordion panel content is displayed. Each accordion header contains an ARIA label, and the header buttons have an ARIA-control property set that points to the unique id of the panel it controls. When focus is on the Accordion header there is a prominent style change to the border as well as background of the accordion header.
```
janchild, line 29:
```suggestion
When an accordion panel is collapsed the content is hidden, therefore the ARIA state `aria-expanded="false"` advises users of assistive technologies that the accordion panel is not visible. When the accordion header is expanded the ARIA state changes to `aria-expanded="true"` and the accordion panel content is displayed. Each accordion header contains an ARIA label, and the header buttons have an ARIA-control property set that points to the unique id of the panel it controls. When focus is on the accordion header there is a prominent style change to the border as well as background of the accordion header.
```

PR #1120, opened 2020-04-20
fix(checkbox): usage update

### LINKED ISSUE #1067: docs: Update Checkbox to follow new component template
Component template details here: https://www.notion.so/carbondesignsystem/Component-template-working-doc-d0d04df0f71c4c64a45c01834ee90336

Find (and add) specific action items for components here: https://www.notion.so/carbondesignsystem/7a1a698e3ad544b1a4180e29febc0cd2?v=bcad0aa949cb4ce1931a71b3be24a771

laurenmrice:
Checkbox usage content update:
-New images
-New docs

Closes #1067


src/pages/components/checkbox/usage.mdx
```diff
@@ -1,33 +1,48 @@
 ---
 title: Checkbox
-description: Checkboxes are used for a list of options where the user may select multiple options, including all or none.
+description: Checkboxes are used when there are one or more items to select in a list.
 tabs: ["Code", "Usage", "Style", "Accessibility"]
 ---
 
+<PageDescription>
+
+Checkboxes are used when there are one or more items to select in a list.
```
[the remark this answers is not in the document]
laurenmrice: ooh. yeah I think they should be different, thank you!

src/pages/components/checkbox/usage.mdx
```diff
@@ -1,33 +1,48 @@
...
 
-The checkbox control allows for three states: **selected**, **unselected**, and **indeterminate**. The indeterminate state comes into play when the checkbox contains a sublist of selections, some of which are selected, and some unselected.
+#### Forms
 
-#### Click target
+Can be used in forms on a full page, in modals, or on side panels.
+
+#### Filtering and batch action
+
+Used to filter data either on a page, in a menu, or within a component. Checkboxes are found in our [Multiselect dropdown](components/dropdown/code) to select multiple items and in [Data table](components/data-table/code) for batch editing purposes.
```
joshblack, line 37: Let me know if this would be confusing or not, but on the development side a `Checkbox` component would correspond to this definition from WAI ARIA:

> Dual-state: The most common type of checkbox, it allows the user to toggle between two choices -- checked and not checked.

For selection in a component, like for multi-select, they are styled to show that an option is selected but unfortunately we don't use `Checkbox` in this situation.

The difference is that a checkbox corresponds to the following input in HTML:

```html
<label for="checkbox">This is a checkbox</label>
<input id="checkbox" type="checkbox">
```

Since `input`'s are interactive, they unfortunately aren't able to be used in things like multi-select.

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+
+<Row>
+<Column colLg={8}>
+
+![The default state for checkboxes is unchecked.](images/checkbox-usage-3.png)
+
+</Column>
+</Row>
+
+1. **Heading (optional):** Communicates what needs to be selected below.
```
joshblack, line 88: I think we might need to flag this as required potentially, but I think it's right to say that this doesn't have to present itself visually.

In code this group might look like:

```html
<h3 id="group-label">Languages</h3>
<div role="group" aria-labelledby="group-label">
  <ul>
    <li>
      <input id="option-1" type="checkbox" />
      <label for="option-1">Node.js</label>
    </li>
    <li>
      <input id="option-2" type="checkbox" />
      <label for="option-2">Go</label>
    </li>
    <li>
      <input id="option-3" type="checkbox" />
      <label for="option-3">Java</label>
    </li>
  </ul>
</div>
```

And as a result we do need a label for the group, at the very least in code but ideally presented to both user groups.

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+![The default state for checkboxes is unchecked.](images/checkbox-usage-8.png)
+
+</Column>
+</Row>
+
+### Interactions
+
+#### Click target
+
+Users can select an item by clicking on the checkbox control directly or by clicking on the checkbox label. Having both regions interactive creates a more accessible click target.
```
joshblack, line 162: I think the click target area for this might be off by a bit, unfortunately I think the space around the checkbox is margin currently:

image

And so wouldn't be interactive 😞 

 I think the click target for the text would be the line-height (24px?). The image could 100% have that exactly, just couldn't tell 😅 

Similarly, if we wanted to do guidance for how to interact with keyboard/SR we could use the same copy from accordion since I believe it's space/enter along with ctrl+option+space!

src/pages/components/checkbox/usage.mdx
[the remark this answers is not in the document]
janchild: I'm working on something for you, @laurenmrice !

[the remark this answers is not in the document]
laurenmrice: thank you!!

src/pages/components/checkbox/usage.mdx
laurenmrice: OH okay, so the click target for the control is just the icon bounds only (not including the margin). And the click target for the label would be the line height.

If you think it would be helpful to add the keyboard guidance we can! Is it the exact same copy from Accordion or does it need to be tweaked?

src/pages/components/checkbox/usage.mdx
laurenmrice: so this is not a checkbox situation for the multiselect dropdown. is that also true for the selectable rows in a data table?

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+
+<Row>
+<Column colLg={8}>
+
+![The default state for checkboxes is unchecked.](images/checkbox-usage-4.png)
+
+</Column>
+</Row>
+
+<Caption>Vertically stacked versus horizontal alignment.</Caption>
```
janchild, line 104:
```suggestion
<Caption>Vertically stacked versus horizontal alignment</Caption>
```

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+
+### Placement
+
+The checkbox component is often used in forms. Forms can be placed on a full page, in a modal or in a side panel. A checkbox can also be used for agreeing to terms and conditions or to filter information.
```
janchild, line 108:
```suggestion
The checkbox component is often used in forms. Forms can be placed on a full page, in a modal, or in a side panel. A checkbox can also be used for agreeing to terms and conditions or to filter information.
```

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+<Row>
+<Column colLg={8}>
+
+![The default state for checkboxes is unchecked.](images/checkbox-usage-6.png)
+
+</Column>
+</Row>
 
-Whereas radio buttons represent a group of _mutually exclusive_ choices, users can select one or more checkboxes from a group. In use cases where only one selection of a group is allowed, use the radio button component instead of the checkbox.
+<Caption>Spacing between a checkbox and other components in a form.</Caption>
```
janchild, line 130:
```suggestion
<Caption>Spacing between a checkbox and other components in a form</Caption>
```

src/pages/components/checkbox/usage.mdx
laurenmrice: In code, if there was one checkbox present would you need a heading for it in code? (ex: terms and conditions). if not, then maybe we just need to clarify there that headings for groups of checkboxes is required.

src/pages/components/checkbox/usage.mdx
```diff
@@ -37,35 +52,191 @@ The default view of a set of checkboxes is having no option selected.
...
+
+<Row>
+<Column colLg={8}>
+
+![The default state for checkboxes is unchecked.](images/checkbox-usage-5.png)
+
+</Column>
+</Row>
+
+Checkboxes in a form should be placed at least 32px (`layout-03`) below or before the next component. Having 24px or 16px can also be used when space is more restricted or if the form is more complex.
```
janchild, line 118:
```suggestion
Checkboxes in a form should be placed at least 32px (`layout-03`) below or before the next component. Spacing of 24px or 16px can also be used when space is more restricted or if the form is more complex.
```

src/pages/components/checkbox/usage.mdx
joshblack: I think it's the same, yeah! The only change would be the focus outline placement.

src/pages/components/checkbox/usage.mdx
joshblack: @laurenmrice good question, I don't know if we would say heading necessarily but we would need a label. If you look at that markup, we would say that the label for the checkbox is the `<label>` element, for example:

```html
<input id="option-3" type="checkbox" />
<label for="option-3">Java</label>
```

We would say that the label for the checkbox is `Java`.

The group also has a label, but instead, it uses `aria-labelledby` instead of the `<label>` element. The `aria-labelledby` attribute points to the heading `<h3>` tag and treats its text as the label for the group.

src/pages/components/checkbox/usage.mdx
joshblack: I believe for rows in data table we still use a checkbox! This is definitely controversial but seems fair to have it in here since there is no official solution for row selection in a table right now so really nothing works 🤦 lol

src/pages/components/checkbox/usage.mdx
laurenmrice: Okay, For this we can say something about how visually its not required unless there is a grouping of two or more checkboxes and also say that in code it needs to be required even if its just one checkbox and can reference the code tab for my information.

laurenmrice: Is it okay that we are labeling the anatomy of the checkbox as 1. Heading. 2. Checkbox and 3. Label ? If in fact `<label>` is referencing the "heading". In the react demo and on the component switcher demo the text is showing how I am currently labeling them here.

joshblack: @laurenmrice I think it breaks down into:

- A checkbox input requires a label
- A group of checkbox inputs requires a grouping label to describe the group of checkbox inputs
- A label is always required for a checkbox input and group of checkboxes, but in rare cases it does not need to be visible. To understand how to label something without it showing up visually, see ___.


image

src/pages/components/checkbox/usage.mdx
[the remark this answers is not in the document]
laurenmrice: fixed 👍🏻

src/pages/components/checkbox/usage.mdx
laurenmrice: In the newest build, I changed all the content on the page to reflect this. ^

src/pages/components/checkbox/usage.mdx
```diff
@@ -1,71 +1,261 @@
 ---
 title: Checkbox
-description: Checkboxes are used for a list of options where the user may select multiple options, including all or none.
+description: Checkboxes are used when there are multiple items to select in a list. Users can select zero, one, or multiple items.
```
emyarod, line 3: should we just say users can select any number of items?

src/pages/components/checkbox/usage.mdx
```diff
@@ -1,71 +1,261 @@
...
+
+<Row>
+<Column colLg={8}>
+
+![Placement of checkboxes in different UI scenarios.](images/checkbox-usage-5.png)
+
+</Column>
+</Row>
+
+Checkboxes in a form should be placed at least 32px (`layout-03`) below or before the next component. Spacing of 24px or 16px can also be used when space is more restricted or if the form is more complex.
```
emyarod, line 118: do we want to mention the tokens for 24px and 16px here too?

src/pages/components/checkbox/usage.mdx
```diff
@@ -1,71 +1,261 @@
...
+<Column colLg={8}>
+
+![Click targets for checkbox.](images/checkbox-usage-9.png)
+
+</Column>
+</Row>
+
+#### Keyboard
+
+Users can trigger a state change by pressing the `Enter` or `Spacebar` key while the checkbox input has focus. For additional keyboard interactions, see the [accessibility tab](https://www.carbondesignsystem.com/components/checkbox/accessibility).
```
emyarod, line 224: is `Enter` to select/deselect expected behavior? we currently don't have this and it's not default behavior for checkboxes

laurenmrice: Okay i will take that out

emyarod: by default, space bar selects and deselects but Enter doesn't, and our components follow that pattern

emyarod: I didn't see your edit originally, but I don't think JAWS and NVDA change the interaction pattern (not entirely sure on that though since I don't have access to either)

laurenmrice: @dakahn do you have any insight on this?

dakahn: Space (not enter) to select and deselect is expected default behavior. Unless we have really really good reason to we defer to the [WAI-Aria Authoring Practices Guide](https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190814/examples/checkbox/checkbox-1/checkbox-1.html) for keyboard interaction patterns.

PR #1129, opened 2020-04-24
docs(modal): usage content

aagonzales:
Closes https://github.com/carbon-design-system/carbon-website/issues/1068

Updated modal content to reflect the new template and image guidelines.


asudoh:
Nice updates @aagonzales - From my quick look one naïve question; Would it be worth mentioning when to use "modal for notifying urgent information" vs. inline/toast notification?

aagonzales:
@jeanservaas ok I've updated the images for modal. I'll update the accordion one separately since I need to do a few other tweaks to the accordion page. I think the divider line color and the button issue would be separate from this PR. 

These are the two issues Shixie opened on that topic. I think we decided stacking buttons was allowed at smaller break points we just need to add that to the code first before documenting for we don't get in trouble. I'm not sure what we want to do about the divider rule between buttons.
- https://github.com/carbon-design-system/carbon/issues/5638
- https://github.com/carbon-design-system/carbon/issues/5636

aagonzales:
@asudoh I added a section under "related" to help and explain the difference between notification and modal.

src/pages/components/modal/usage.mdx
```diff
@@ -165,13 +165,13 @@ For further content guidance, see Carbon's [content guidelines](/guidelines/cont
...
+Modals are triggered as a result of a user’s action and are not system generated. Common components that can trigger a modal include, button, link, or icon. On a keyboard, selecting `Enter` or `Space` should launch the modal.
 
 ### Focus
 
-Once the modal is open, set the initial focus to the first location that accepts user input. Focus should then remain trapped in the dialog until it is closed.
+Once the modal is open, set the initial focus to the first location that accepts user input. For example, if the modal contains a form then the focus should automatically be set to the first field when opened. If it is a transactional modal without form inputs in the body section then the first focus should be on the primary button.
 
-For example, if the modal contains a form then the focus should automatically be set to the first field when opened. If it is a transactional modal without form inputs in the body section then the first focus should be on the primary button.
+Focus should then remain trapped in the dialog until it is closed. When navigating by keyboard `Tab` and `Shift-Tab` do not move the focus outside of the modal.
```
janchild, line 174:
```suggestion
Focus should then remain trapped in the dialog until it is closed. When navigating by keyboard, `Tab` and `Shift-Tab` do not move the focus outside of the modal.
```

src/pages/components/modal/usage.mdx
```diff
@@ -330,7 +330,18 @@ There are three buttons in the progress modal footer: Cancel, Previous, Next. Ea
 
 ## Related
 
+#### Modal vs. notification
+
+Modals are highly disruptive and block tasks until dismissed by the user. A modal should only be used as a form of notification if it provides the user with critical information immediately related to their task. For non-critical messaging considering using a toast or inline notification instead of a modal. Toast and inline notifications provide users with non-disruptive feedback or the status of an action without breaking their workflow.
```
janchild, line 335:
```suggestion
Modals are highly disruptive and block tasks until dismissed by the user. A modal should only be used as a form of notification if it provides the user with critical information immediately related to their task. For non-critical messaging, consider using a toast or inline notification instead of a modal. Toast and inline notifications provide users with non-disruptive feedback or the status of an action without breaking their workflow.
```

PR #1126, opened 2020-04-24
feat(code): add code tab content to modal

asudoh:
#### Changelog

**New**

- Updated code doc for modal with technical usage instructions. The team should discuss the unified content style in this area, but for this PR, would like to focus on:
  - Improving the answers on "how to use feature X technically" questions
  - Getting the look and feel right

PR #1134, opened 2020-04-27
docs(accordion): add initial code docs for accordion

joshblack:
Adds in the initial code docs for accordion for feedback 👀 This is the first iteration for accordion docs and we're still trying to nail down format, how we talk about things, what is appropriate for this page versus not, etc. All questions/feedback welcome!

Some of our remaining questions:

- What would be a good page description for these pages? (If any)
- What should live in the TOC at the top of the page? Currently, we have:
  - Usage
    - Common usage scenarios for this component
  - Component API
    - Reference documentation for a specific component
  - Examples
    - Links to external examples that are not appropriate for this page
  - FAQ
    - Common questions either from the above documentation or from Slack
- What kinds of examples make sense on this page versus their own dedicated example on codesandbox?
- How should we introduce/talk about these components in a voice and tone that's consistent with the site?

Some enhancements that we're expecting to come for the Component Demo:

- Instead of relying on the full preview, use an inline variant that:
  - Removes links
  - Removes theme switcher
  - Removes knobs
- Pin content to top-left and have a block layout instead of an inline layout
- Hot Module Replacement support for code snippets

#### Changelog

**New**

**Changed**

- Update accordion/code.mdx page with additional reference documentation and examples

**Removed**


alisonjoseph:
Quick question, maybe for @janchild or @aagonzales , is the main component demo moving from the code page to the usage page? and should that be the first tab now, or will we reorganize tabs across the board in a different PR? Just want to make sure we are syncing on the general layout between these two PRs (https://github.com/carbon-design-system/carbon-website/pull/1126) and future components.

janchild:
> Quick question, maybe for @janchild or @aagonzales , is the main component demo moving from the code page to the usage page? and should that be the first tab now, or will we reorganize tabs across the board in a different PR? Just want to make sure we are syncing on the general layout between these two PRs (#1126) and future components.

@alisonjoseph I think there's still a bit of discussion needed here and so that would be part of a different PR.

aagonzales:
@alisonjoseph @janchild Oh yeah I keep meaning to open an issue for that, I think we would do that change all at once so its consistent. And then once it's finished these code pages would use the more simply component demo.

The only catch with that is right now on all of the code tabs the only thing there is the component demo so if we move it those pages will be empty until we get the dev docs done. We could probably design some sort of placeholder for those pages that link out to the other repos or something.

aagonzales:
I think for the most part this seems like a good structure. 
- I like the list/tables it makes the content is to parse and act on.
- We'll probably want to do some clean-up on the examples demos. Ideally the toggle button for skeleton would go above the accordion and not beside it. And the content with the accordion demo should probably read "Header A" and then "Panel A" they both say "Panel A" right now. 
- Akira had an install section, if that something we want on all pages or is this handled in the Getting started section? https://carbon-website-git-fork-asudoh-modal-code.carbon-design-system.now.sh/components/modal/code#installation
- Accordion is pretty basic on usage docs so I'm curious what else could go into this section. For Akira's modal usage section is much more robust. We should also probably make a rule of when you should show a demo and when just showing a code snippet if ok. 
- We'll probably need to look at the "Note" and info notification and decide what we want to do with these and when its appropriate to use them.  I'm not sure all of Akira's are needed inside of a info notifications. 
image
vs
image

joshblack:
Thanks all for the feedback! 🙏 

For specific points:

> We'll probably want to do some clean-up on the examples demos.

Agreed! I think this will go alongside the inline demo where things get pinned top left instead of center center.

> Akira had an install section, if that something we want on all pages or is this handled in the Getting started section

Great point, we were talking about it and it seems like there are a couple of perspectives on it. I tried to capture "top-level" topics over in this Notion note: https://www.notion.so/carbondesignsystem/Top-level-topics-4ad32e28a05f4d21a25bc4182e124a73

It might make sense for install topics to live at this top-level section, otherwise we end up with the same install instructions on each page.

There is a detail that may be helpful for people to know from that install section that would be unique to each page, specifically how to import only that component's style. Maybe it'd be helpful to have a setup section or something similar? Or optimization?

Similarly, I love that the install section listed out all of the components available. that seems huge.

> Accordion is pretty basic on usage docs so I'm curious what else could go into this section.

I think a lot of this is because accordion usage is simpler than Modal/Dialogs. There are a couple more props we could expose but I think what is on the page covers most of the usage for folks.

> We'll probably need to look at the "Note" and info notification and decide what we want to do with these and when its appropriate to use them

Makes sense! Would it make sense for me to update to the info style or bring this out into a separate issue?

aagonzales:
> Accordion is pretty basic on usage docs so I'm curious what else could go into this section.

Oh yeah I should have clarified I meant this section in general, not specifically related to accordion. 


> We'll probably need to look at the "Note" and info notification and decide what we want to do with these and when its appropriate to use them
> Makes sense! Would it make sense for me to update to the info style or bring this out into a separate issue?

Maybe make this a separate issue. I think the info notification is actually too much or at least we should limit how much it's used. I'm wondering if there's another style for this or if just having an italic note is fine. I'm sure Jeannie and Mike have thoughts on it.

joshblack:
Sounds good! On it 👍

package.json
```diff
@@ -103,5 +104,12 @@
     "*.{css,scss,json,html,yaml,md,mdx}": [
       "prettier --write"
     ]
+  },
+  "prettier": {
```
vpicone, line 108: We use eslint-plugin-prettier to format things, running eslint on a file should take care of the formatting/js issues in one go.

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+  content and can choose to expand that content.
+tabs: ['Code', 'Usage', 'Style', 'Accessibility']
 ---
 
+import usage from './snippets/usage';
+import skeleton from './snippets/skeleton';
+
+<PageDescription>
+
+Learn how to build an accordion in React.
```
vpicone, line 15: Would it be better to explain what an Accordion is briefly here? Thinking about other pages getting a little repetitive: "Learn how to build a[Dropdown,Select,DataTable] in React"

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+Additional props passed into `Accordion` will be forwarded along to the
+underlying accordion container.
+
+### Accordion align
+
+In rare cases, you may need to specify the alignment of the icon in the
+accordion. You can use the `align` prop to specify the side where the icon
+should be placed.
+
+_Note: This prop must not be used to create a tree view or set of nested
```
vpicone, line 72: These note's might be better suited as [inline notifications](https://gatsby-theme-carbon.now.sh/components/InlineNotification). What do you think @aagonzales?

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+You can build an accordion using a combination of the `Accordion` and
+`AccordionItem` components. The `Accordion` components acceps a list of
+`AccordionItem` components as `children` and each `AccordionItem` is responsible
+for displaying the accordion's heading and panel content.
+
+You can configure the accordion item's heading using the `title` prop.
+Everything you pass in as a child of `AccordionItem` will be rendered in the
+accordion's panel.
+
+<ComponentDemo links={{}}>{usage}</ComponentDemo>
```
jnm2377, line 38: Just wondering why we don't want to have the knob container here to show the interaction of the props? I thought that was kind of the point of the knobs in the demo? So users can see the changes in action.

src/pages/components/accordion/snippets/usage.js
```diff
@@ -0,0 +1,14 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export default `
+<Accordion>
+  <AccordionItem title="Panel A">Panel A</AccordionItem>
+  <AccordionItem title="Panel B">Panel B</AccordionItem>
+  <AccordionItem title="Panel C">Panel C</AccordionItem>
+</Accordion>
+`;
```
jnm2377, line 14: In my component demo PRs design had suggested having the text be something more meaningful to the accordion example, which I kind of liked as well. 
Screen Shot 2020-04-30 at 1 59 50 PM

package.json
joshblack: Yeah, was running into issues with format-on-save so added that in lol. This would also correspond to the `format` command too that runs on commit for the file types listed, right? Or are the settings shared somewhere for formatting those files?

src/pages/components/accordion/code.mdx
joshblack: @vpicone yeah, there was some chatter about it below where we talked about this.

src/pages/components/accordion/code.mdx
joshblack: @jnm2377 great question, I think knobs, in general, have presented a challenge for us on the documentation side. Tried to share a little bit more about this over in https://github.com/carbon-design-system/carbon/issues/5640#issuecomment-601255445

Some of this is very much so storybook specific (like copy/paste-ability), other bits are kind of a consequence of the format. The biggest shortcoming is that we would like to offer inline snippets/explanations for props and how they should be used either in isolation or in conjunction with other props or scenarios.

src/pages/components/accordion/snippets/usage.js
joshblack: Agreed!

src/pages/components/accordion/code.mdx
joshblack: Agreed! Don't know what's better though. Happy to remove.

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+
+The `className` prop passed into `AccordionSkeleton` will be forwarded along to
+the underlying skeleton container. This is useful for specifying a custom class
+name for layout.
+
+```jsx
+<AccordionSkeleton className="custom-class" open count={3} />
+```
+
+## FAQ
```
joshblack, line 164:
```suggestion
## Reference
```

package.json
joshblack: Also forgot to mention, it specifically was for .mdx files to be formatted with my editor.

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+
+Learn how to build an accordion in React.
+
+</PageDescription>
+
+<AnchorLinks>
+
+<AnchorLink>Usage</AnchorLink>
+<AnchorLink>Component API</AnchorLink>
+<AnchorLink>FAQ</AnchorLink>
```
joshblack, line 23:
```suggestion
<AnchorLink>Reference</AnchorLink>
```

src/pages/components/accordion/code.mdx
janchild, line 15:
```suggestion
How to build an accordion using React.
```

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
+<AnchorLink>Usage</AnchorLink>
+<AnchorLink>Component API</AnchorLink>
+<AnchorLink>Reference</AnchorLink>
+
+</AnchorLinks>
+
+## Usage
+
+You can build an accordion using a combination of the `Accordion` and
+`AccordionItem` components. The `Accordion` components acceps a list of
```
janchild, line 30:
```suggestion
`AccordionItem` components. The `Accordion` components accepts a list of
```

src/pages/components/accordion/code.mdx
```diff
@@ -1,22 +1,177 @@
...
-`}</ComponentDemo>
+```
+
+### AccordionItem title
+
+You can use the `title` prop to specify the accordion item's heading. The
+default behavior is to pass in a `string` as the title, however you can also
+pass in a `node`.
+
+_Note: a custom title prop that renders an interactive element is considered an
```
janchild, line 131:
```suggestion
_Note: A custom title prop that renders an interactive element is considered an
```

src/pages/components/accordion/code.mdx
joshblack: Seems like Jan offered a suggestion below 👀 added

package.json
joshblack: Let me know if it makes sense to have a shared prettier config either here or in `prettier.config.js` for this use-case (that can also be read by `eslint-plugin-prettier`) 👀

joshblack:
Gonna close this out and re-open with less merge conflicts 👀

PR #1142, opened 2020-04-30
chore: update theme and fix dropdown

vpicone:


vpicone:
@sstrubberg @andreancardona should be fixed now

PR #1160, opened 2020-05-05
chore(project): add prettier and format commands, add ci-check command

joshblack:
Noticed this over on the Accordion PR where it seemed like prettier may not be getting applied to different filetype consistently. Similarly, there was no check to see if a file had been formatted by prettier or not.

This PR adds in support for a CI check workflow for checking prettier usage alongside linting, it also sets a shared prettier config that is applied whenever the format and format diff commands are run. The final step is a new `ci-check` command that is run with the associated GitHub Action workflow that has been added under `.github/workflows/ci.yml` that is run on each Pull Request.

#### Changelog

**New**

- GitHub Workflow that runs on master and pull requests to check if prettier has been applied and lint passes. In the future might make sense for tests to be added, as well.
- Add `ci-check` command
- Add `format:diff` command
- Add `prettier` config to `package.json

**Changed**

- Applied prettier to files in the project

**Removed**


vpicone:
Hey thanks for this. Hopefully we can figure out what was causing the inconsistent application issues you were experiencing. Just as background, we use eslint (via eslint-plugin-prettier) to format js and prettier for everything else. When you commit, commitlint will run eslint if it's js and prettier if it's not. 

It looks like your PR includes changes to `js` files, which should be under the purview of eslint, not prettier. Do you know maybe why those files got impacted?

joshblack:
Hey! To clarify, this PR creates a prettier config that is shared across prettier setups and doesn't substitute what already exists 😄 

This config mirrors what we have over in the monorepo and the config itself is used by the `prettier` cli and is picked up by `eslint-plugin-prettier` as the configuration to apply.

Changes to the files in the diff were caught by `yarn format:diff` and come up as a result of the CI check for files that aren't formatted by this rule. The PR ran both `yarn format` and `yarn lint` to format files. The changes in the config are what is causing the changes in JS files 👍

vpicone:
@joshblack nice, that makes sense. We lean on the default prettier configuration for the theme, carbon, idl, and brand sites but it'd be nice if all of the Carbon projects used the same config. A couple things about the implementation here.

- Unlike lint-staged, I like how this would catch `edit in github` edits which is really nice. But, I think having both lint-staged and the action seems like overkill. Not sure what the advantage of lint-staged would be at that point what do you think?

- If we were going to move lint-staged behavior into actions, I'd rather we just format and fix rather than diffing/failing the build.

- I think we don't need the build step here, since we're just interacting with the source. That should shave off like 15 min haha.

joshblack:
> But, I think having both lint-staged and the action seems like overkill.

I think it can still be helpful to have both. `lint-staged` is super convenient and the action is helpful for catching files that may have been forgotten to be formatted (like if you have pass in `--no-verify` to a commit).

> If we were going to move lint-staged behavior into actions, I'd rather we just format and fix rather than diffing/failing the build.

I think we looked into doing an autofix + commit over on the monorepo, but unfortunately, I think the token ends up not having permissions to do the commit on a PR from a fork but I could be misremembering what the issue was.

> I think we don't need the build step here, since we're just interacting with the source. That should shave off like 15 min haha.

Makes sense. Definitely happy to remove.

---

Going to revert my PR back from a draft, not sure how it was converted to one in the first place 🤔

vpicone:
@joshblack I can take a look again when I get back, I think I'd rather just use the latest prettier defaults for everything rather than adding a custom config to all the sites.

Our image tool has permissions to commit from an action so not sure why prettier wouldn't be able to.

joshblack:
> Our image tool has permissions to commit from an action so not sure why our prettier/format action wouldn't be able to.

@vpicone I believe the calibre actions might run into the same limitations for PRs made from forks based on the permission table for the token defined [here](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token)

It seems like this detail was brought up over in: https://github.com/calibreapp/image-actions/issues/53

vpicone:
Sounds good, so do you think it’s worth changing the prettier settings from the default for this site away from the other four sites? They have different sorts of content being styled than the monorepo and I’d like to keep the default unless there’s a solid reason to break them up otherwise. 

Otherwise for modifications to Js prettier settings we’d want to modify those as part of the eslint config rather than split settings between to config files.

joshblack:
> If we can't figure out how to commit the formatting, I'd rather not bring in the build check. 

Happy to shift the formatting to a commit-after-merge strategy like we do for the generated files in Carbon 👍 

> IRT the config, do you think it’s worth changing the prettier settings from the default for this site away from the other four sites? I’d like to keep the prettier default unless there’s a solid reason to break them up otherwise

I think the hope right now is to have a single prettier config shared across the file types that we run prettier against. For example, in the current setup the wesbos config will set singleQuotes to true for JS files, but in MD/MDX files with JS code blocks the default settings for prettier will use double quotes. Would be great to have this stuff consistent 👍

vpicone:
@joshblack that makes sense. What do you think about using the prettier defaults for everything across all projects?

joshblack:
@vpicone I think there are some options that would be nice to have, namely single quote and prose wrap. I also don't think all projects necessarily have to use the same setup if it doesn't make sense 👍 Would it help if we published a package for this?

vpicone:
@joshblack I think so, I know you've been a little hesitant on publishing/maintain our eslint config before. Would be worth digging into especially if we want to set a base config for community components.

What's the motivation for prose wrap? It'd be a bummer to add prettier config just so we can have single quotes, maybe if it were to move everything to template literals haha

joshblack:
Closing this out after talking to vince with some action items:

- Wait till `prettier-config-carbon` is published as an RC
- Create one PR that applies `prettier-config-carbon` to the files
- Defer on GitHub actions for auto-formatting unless folks are interested

PR #1163, opened 2020-05-06
Disabled image swap

connor-leech:
Our guidance suggests disabled elements should not be interactive. Swapped out the disabled state image showing a tooltip on a disabled primary button.

Also snuck some hyphen removal in 😉.


connor-leech:
@alisonjoseph Is the reformatting of the `react-docgen.json` file an issue?

alisonjoseph:
No idea what that file is, maybe @vpicone can help 😄

jnm2377:
@alisonjoseph that file is what we use for the component demo knob data for all of the components.

jnm2377:
Looks like the formatting just changed the indentation. The component demos all appear to be working correctly, so it should be fine. Although idk why the indentation changed, it looks weirder to me this way. 🤷‍♀️

connor-leech:
@jnm2377 I'm guessing it was Prettier

vpicone:
hmmmm, I wonder why it got ran here @connor-leech. Did you maybe edit something in that file by chance? Either way you can run `git checkout master -- ./src/data/docgen/react-docgen.json` from the package root to undo it in this PR. If we need to run prettier on it, we wanna do that in it's own issue.

PR #1196, opened 2020-05-13
remove unnecessary dataviz tutorials

theiliad:
Removing dated and unnecessary dataviz getting started tutorials.


alisonjoseph:
I think you will need to update the left nav file also since the url for the first page won't be /vanilla anymore. (we should also make sure no other pages are referencing that link)

theiliad:
> I think you will need to update the left nav file also since the url for the first page won't be /vanilla anymore. (we should also make sure no other pages are referencing that link)

Good point, just pushed that. And running just a simple file search I don't think there are any other links to the page

theiliad:
Hi team, ready to merge?

theiliad:
just a heads up, we need this PR out asap since the info that's currently on the site is dated and irrelevant

PR #1217, opened 2020-05-22
docs(notification): add notification code docs

dakahn:
Updates Notification code docs according to (my interpretation) of our template.


dakahn:
Removing WIP label since it seems like it all built fine and dandy

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+}
+```
+
+## Component API
+
+### Inline Notification props
+
+| Prop                  | Type       | Required | Default               | Description                                                            |
+| --------------------- | ---------- | -------- | --------------------- | ---------------------------------------------------------------------- |
+| actions               | `node`     | -        | -                     | Pass in the action nodes that will be rendered inside the notification |
```
joshblack, line 78: Make sure all dashes are en-dashes

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
-  <div>
+<PageDescription>
+
+How to build a notification using React.
+
+</PageDescription>
+
+<AnchorLinks>
+
+<AnchorLink>Overview</AnchorLink>
```
joshblack, line 15: would it be helpful to add anchor links for toast notification and inline notification?

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+
+</AnchorLinks>
+
+## Overview
+
+Carbon Notifications come in two variants: toast, and inline. They are created by using the `ToastNotification` or `InlineNotification` components.
+
+## Toast notification
+
+You can use the `ToastNotification` component to display a non-modal, time-based short message that appears at the bottom or the top of the screen.
```
joshblack, line 27: Would this be a `non-modal dialog`?

Also, just realized that we don't provide any kind of layout for where this component should be rendered 🤔

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+
+```jsx
+import { ToastNotification } from "carbon-components-react";
+
+function MyComponent() {
+  return (
+    <ToastNotification
+      caption="00:00:00 AM"
+      iconDescription="describes the close button"
```
joshblack, line 36: Maybe: "Dismiss notification" or "Close notification"? (Totally get that this is from the knobs, just was thinking what might be more helpful in the example for folks since we do describe the prop down below)

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+      subtitle={
```
joshblack, line 37: Would we want to demonstrate the text-based prop instead of the custom node use-case? Trying to think about the simplest example here, we could always show this example as a custom node variant later on too.

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+## Toast notification
+
+You can use the `ToastNotification` component to display a non-modal, time-based short message that appears at the bottom or the top of the screen.
+
+```jsx
+import { ToastNotification } from "carbon-components-react";
+
+function MyComponent() {
+  return (
+    <ToastNotification
```
joshblack, line 34: Would we want to talk about the different notification kinds, as well? Not sure where the best place is.

Similarly, would we want to mention `lowContrast` and how to use that for these components?

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+| iconDescription       | `string`   | `true`   | `closes notification` | Provide screen reader accessible description for close icon            |
+| statusIconDescription | `string`   | –        | –                     | Provide screen reader accessible description for status icon           |
+| notificationType      | `string`   | –        | `inline`              | Specify the type of the button                                         |
+| hideCloseButton       | `bool`     | –        | `false`               | Optional prop to disable the close button                              |
+
+## Inline notifications with action button
+
+An inline notification optionally can have a ghost button that allows the user to take further action.
+
+```jsx
```
joshblack, line 96: Seems like this snippet got all lower-cased (if you were using vim, I totally have had this happened before lol)

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+      }
+      title="notification title"
+    />
+  );
+}
+```
+
+### Notification action button props
+
+| prop      | type       | required | default | description                                   |
```
joshblack, line 118: Style nit: uppercase table titles

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,142 @@ description: Notifications are messages that communicate information to the user
...
+| --------- | ---------- | -------- | ------- | --------------------------------------------- |
+| classname | `string`   | –        | –       | class name to be applied to the action button |
+| children  | `node`     | -        | –       | the content of teh notification button        |
```
joshblack, line 121:
```suggestion
| children  | `node`     | -        | –       | the content of the notification button        |
```

src/pages/components/notification/code.mdx
dakahn: Like we just say "render at the top" but don't describe how to do that?

src/pages/components/notification/code.mdx
dakahn: Where do we describe the types of notifications? I might be missing it. I'll wing it for now but will deffo need a content pass on whatever I come up with. 👍

src/pages/components/notification/code.mdx
```diff
@@ -4,66 +4,132 @@ description: Notifications are messages that communicate information to the user
...
-        "https://angular.carbondesignsystem.com/?path=/story/components-notification--basic",
-      Vue:
-        "http://vue.carbondesignsystem.com/?path=/story/components-cvinlinenotification--default",
-      Vanilla: "https://the-carbon-components.netlify.com/?nav=notification",
-    }}
-  >{`
-  <div>
+<PageDescription>
+
+How to build a notification using React.
```
joshblack, line 9: I think this needs to include the link for other frameworks that was added to the template 👀

janchild: Yes, just this one sentence but no link, so the complete description would be:

> How to build a [component name] using React. For code usage with other frameworks, please follow the links in the live demo on the usage tab.

dakahn:
Updates added! Should be good to go :surfer:

janchild:
@dakahn , there is one H2 that's not currently in the anchor links: **Inline notifications with action button**. And it looks like it should probably be positioned after the H2 **Inline notifications** section, with the Component API and the Feedback H2s following behind. Is that right? I might be missing something here... 

So I'm thinking it should be:
- Toast notification
- Inline notification
- Inline notifications with action button
- Component API
- Feedback 

Thanks

dakahn:
@janchild re organized the H2s and added the Overview anchor back in :+1:

PR #1248, opened 2020-06-02
fix(icon-library): update dropdown max-height

jnm2377:
Rel: https://github.com/carbon-design-system/design-language-website/issues/560

### DIFF
--- a/plugins/gatsby-theme-carbon-svgs/components/shared/SvgLibrary.module.scss
+++ b/plugins/gatsby-theme-carbon-svgs/components/shared/SvgLibrary.module.scss
@@ -67,6 +67,10 @@
   }
 }
 
+.filter-row :global(.bx--list-box--expanded .bx--list-box__menu) {
+  max-height: 16.5rem;
+}
+
 .dropdown {
   border-bottom: 0;
 }


joshblack:
Seems like this may be a mismatch in code vs guidance, potentially?

cc @laurenmrice noticed you specified this in the new guidance over in: https://www.carbondesignsystem.com/components/dropdown/usage#scrolling-behavior

Do we need to make sure this is updated in code on the system side?

laurenmrice:
-We already do this in code. Currently in code its cropping the 4th option. 
Screen Shot 2020-06-02 at 4 27 14 PM

-In the usage guidance for dropdown, I recommended cropping at the 5th option as a rule since this is being used for product UIs, however this could also depend on the usecase at hand. https://www.carbondesignsystem.com/components/dropdown/usage#scrolling-behavior

-Mike is suggesting to crop at the 6th option in the issue.

I think it is the matter of deciding what our recommendation would be of when to crop. 
@mjabbink what do you think?

joshblack:
@laurenmrice totally makes sense to me. It seems like after we decide on 5 or 6 then we could have a fix put in upstream that the site pulls in versus needing to do an override here. What do you think, @jnm2377? 👀

jnm2377:
@joshblack if we decide to fix it in upstream, I'm fine with closing this PR. I'll wait for design input.

laurenmrice:
@joshblack and @jnm2377 we are going to stick with cropping at the 6th option. I'm going to update the dropdown usage docs to reflect that.

joshblack:
Sounds great, @laurenmrice. Would you mind making an issue for updating this in code? I'll make sure we tackle it this week/next 👍

laurenmrice:
on it! thank you

PR #1265, opened 2020-06-08
Fix example Dropdown data object

danielcolinjames:
When the <Dropdown> component is given an `items` array with each object having a `text` field, no text shows up in the dropdown items—just blank options.

Changing each object in the `items` array to have a `label` field makes them show up in the dropdown. I'm not sure why, or where to even look for further documentation on this, and changing it to `label` was just a wild guess that ended up working.

Example here: https://codesandbox.io/s/quirky-morse-4z9hx

#### Changelog

**Changed**

- Updated structure of the example `items` array passed into the `<Dropdown>` in the example code to make the example work.

### DIFF
--- a/src/pages/components/dropdown/code.mdx
+++ b/src/pages/components/dropdown/code.mdx
@@ -124,19 +124,19 @@ import items from "../../../data/components/dropdown.js";
 const items = [
   {
     id: "option-1",
-    text: "Option 1",
+    label: "Option 1",
   },
   {
     id: "option-2",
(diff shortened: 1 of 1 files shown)


tw15egan:
FYI, you can also get this data working as is, but the example is missing the  `itemToString` prop

`itemToString={item => (item ? item.text : '')}`

https://codesandbox.io/s/autumn-currying-g29jx?file=/index.js


Thanks for taking the time to fix this.

PR #1273, opened 2020-06-09
fix: 1255-space-grid-implmentation

### LINKED ISSUE #1255: fix[spacing]: eliminate gap between live example and code snippet on grid implementation page
These gaps between the live grid example and the code snippet appeared on the grid implementation page after the spacing overhaul, this caused the component to stray from the spec. 

- remove these gaps

andreancardona:
Closes #1255


vpicone:
@joshblack we've been more lax with the template for PR's on the website. Usually the issue gives enough context for scope, especially for a change as small as this.

PR #1345, opened 2020-07-07
chore: tweak initial banner render

vpicone:
Use css instead of JS to show/hide the EJI initialism in the banner. This makes it so we don't have to wait for all of the javascript to load before the page renders the correct content for the screen width.


joshblack:
bump @aledavila when you have a sec!

PR #1382, opened 2020-07-15
Remove leading '$' from bash commands

### LINKED ISSUE #744: CLI Copy Button Problem
## Copy button for the CLI copies the "$"

In the Step 1 of the Carbon Tutorial the copy button for CLI git commands copies the "$" so that when it is pasted into the CLI terminal the command does not work. I did not notice this until it was time to "git remote add upstream https://github.com/carbon-design-system/carbon-tutorial.git" as all the previous entries I had typed in manually.

davidnixon:
Closes #744

Remove the leading `$` from bash commands

#### Changelog

**Changed**

- tutorial and getting started pages

### DIFF
--- a/src/pages/get-started/develop/angular.mdx
+++ b/src/pages/get-started/develop/angular.mdx
@@ -43,10 +43,10 @@ The Angular library is maintained by members of the Carbon community. For suppor
 
 Assuming we're starting with a new @angular/cli project:
 
-```sh
-$ npx @angular/cli new my-project --style=scss
-$ cd my-project
-$ npm i --save carbon-components-angular carbon-components
+```bash
+npx @angular/cli new my-project --style=scss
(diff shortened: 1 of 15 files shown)


davidnixon:
I was irritated by the $ copied to the clipboard when I did the tutorial a few weeks ago so ...

davidnixon:
So for example in production:
https://www.carbondesignsystem.com/get-started/develop/react
copy button puts this on the clipboard including the leading `$`
```sh 
$ yarn add carbon-components-react carbon-components carbon-icons
```
But in preview:
https://carbon-website-aqk4chqe5.vercel.app/get-started/develop/react
copy button puts this on the clipboard (no leading `$`)
```sh
yarn add carbon-components-react carbon-components carbon-icons
```

PR #1394, opened 2020-07-17
docs(colors): add two tokens to color usage tab

laurenmrice:
Adding `decorative-01` and `hover-ui-light` to the color usage tab table.

Related:
https://github.com/carbon-design-system/carbon/pull/6442
https://github.com/carbon-design-system/carbon/issues/2439

### DIFF
--- a/src/data/guidelines/color-tokens.js
+++ b/src/data/guidelines/color-tokens.js
@@ -246,7 +246,7 @@ const colorTokens = {
     },
     '$decorative-01': {
       role: [
-        '"Light" variant elements',
+        '“Light” variant elements',
       ],
       value: {
         white: {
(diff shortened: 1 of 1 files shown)


src/data/guidelines/color-tokens.js
```diff
@@ -902,6 +925,27 @@ const colorTokens = {
         },
       },
     },
+    '$hover-ui-light': {
+      role: ['$“Light” variant hover'],
```
emyarod, line 929:
```suggestion
      role: ['"Light" variant hover'],
```

not sure which kind of quotation marks they're supposed to be but we probably don't need the `$` either right?

laurenmrice: oh woops. yeah. they should be the curly quotes

PR #1400, opened 2020-07-21
chore: release Carbon 10.16

vpicone:
**note:** removed carbon-components and carbon-components-react from dependencies in favor of the resolutions property

### Updating packages
When updating packages, the most important bit is updating the lockfile. Our build step uses that to install packages. If you update something in `package.json` you have to run `yarn` to update that file (and then commit it).

### Update steps
1. manually update the resolutions property in `package.json` to `10.16`
2. run `yarn` to get those updates
3. run `yarn --upgrade-interactive` and select all of the carbon packages and any other dependencies


vpicone:
Hmm something's wrong with copy button. It's broken [upstream in the theme](https://gatsby-theme-carbon.now.sh/getting-started#first-steps) as well.

PR #1407, opened 2020-07-22
feat(data-viz): add color palette component

jnm2377:
Closes https://github.com/carbon-design-system/design-language-website/issues/564

Adds `ColorPalette` component to data viz colors pages. 

### To-do:
- [x] fix style bugs
    - [x] palette colors borders 
    - [x] switcher styles (had to override carbon switcher styles to look like design)
    - [x] palette controls borders (dropdown/switcher where it touches colors)
- [x] fix dropdown selection bug (not showing highlight when selection changes)
- [x] add alert colors palette
- [x] update sequential palette design
- [x] add copy button
- [x] fix copy button feedback z-index


jnm2377:
@jeanservaas still working on things here, but you can start seeing the work/checking the colors are correct.

jnm2377:
@jeanservaas I was still working on the border styling when I pushed last week. It should be updated now. It was a bit tricky bc there were a lot of nuances that borders and outlines have that weren't matching up with the spec.

jnm2377:
@jeanservaas still need to update the sequential palette to the new design, but shouldn't take long

jnm2377:
@jeanservaas I was gonna ask if the responsive breakpoints are correct, bc it seems like they should be changing before those breakpoints. i.e. the specs show that it'll go to one column at the `sm/320px` breakpoint, but it seems like it should happen sooner... same goes for changing from 3 to 2 columns....

jeanservaas:
@jnm2377 sorry, I can't interact with the new preview link.

jeanservaas:
@jnm2377 this looks so good and it's so close.

There is one thing I missed about component on the grid (do not hate me).

Notice to the right, how our component is on the columns, aligned with the type above it. The component on IDL hangs 16px into the gutter on the left and the right of the columns... which is how we want it.

## Comparison: Carbon (rt) / IDL (lft)

image

## On the plus side IT LOOKS GREAT AT MOBILE!

image


## At tablet, you want a 16px margin on either side of component, not 32px

image


## Lrg and XL (the right side will have a 16px margin)

image



The only other little problem I'm having is that I can't see the gallery in my local build, but I'm sure that's a problem on my end. Maybe you could just screen grab it for me or something.

image

src/components/ColorPalette/ColorPalette.js
```diff
@@ -0,0 +1,221 @@
...
+  // SWITCHER STUFF
+  const setTrue = () => {
+    if (type === 'sequential') {
+      setContinuous(true); // for sequential palettes
+    } else {
+      setDark(true); // for all other palettes
+    }
+  };
+
+  const setFalse = () => {
```
vpicone, line 109: Usually the `set[variable]` syntax goes with the the second return value from hooks. Here it sounds like you're setting a variable called `false` to something. When in doubt prefer long names that explain completely what's happening.

src/components/ColorPalette/ColorPalette.js
```diff
@@ -0,0 +1,221 @@
...
+        break;
+      case 5:
+        setColorGroup(fiveColor);
+        break;
+      default:
+    }
+  };
+
+  // SWITCHER STUFF
+  const setTrue = () => {
```
vpicone, line 101: Same as `setFalse` maybe we could combine these functions and pass in a variable to determine the behavior (rather than splitting in two)

src/components/ColorPalette/_color-palette.scss
```diff
@@ -0,0 +1,348 @@
+// MAIN CONTAINER
```
vpicone, line 1: In general, most of the new styles we've added with the theme use css modules. Especially for something as compartmentalized as this it not only prevents class collisions but enforces good practices for not cross-referencing styles between sheets. 

Something like `dark-mode` or `light-text` could very easily be used in a different stylesheet and we'd be impacting those styles.

vpicone (approved):

theiliad:
Thank you @jnm2377 for putting this work in! 🎉

sstrubberg (approved):

theiliad (changes requested):
Couple of points @jeanservaas:
1. Gradients on gauge hasn't been approved (we also probably wouldn't want to recommend those)
image

2. image
Couple of gradients on this page we might not want to recommend
image

vpicone:
@theiliad do you mind opening a separate issue for those so we can merge this?

jeanservaas:
@theiliad yes, let's deal with this in a separate issue. I'll address both of those issues next week but I don't want to block this.

Also @jnm2377 and @vpicone Showing this preview to the design crit I noticed that 

- Color in action section needs an anchor link (my bad)

- that resources link is old, the palette is in the master charts file on the gentling start page — can you just remove the resources section entirely?

Sorry!!

vpicone:
@jnm2377 feel free to merge it after @jeanservaas's things get handled

jnm2377:
@vpicone already pushed a fix, I can't merge tho. It won't let me.

alisonjoseph (approved):

Issue #172, opened 2019-08-26
textarea padding


SpikedKira:
https://www.carbondesignsystem.com/components/text-input/style/
lists left and right padding and height for the text-input, however https://www.carbondesignsystem.com/components/text-input/code/ covers both the input and textarea components. 

1) there is no defined padding token for the textarea
2) the top and bottom padding in the example on the `code` tab are `.6875rem` which isn't defined here: https://www.carbondesignsystem.com/guidelines/spacing

aagonzales:
Task:
- [ ] Add style tab content specifically for text area (currently missing)
- [ ] Investigate top and bottom margin. I believe it should be 12px / `spacing-04`

aagonzales:
Padding top of the textarea matches the padding top of text input. Type within a field is always vertically centered and not controlled by a token.

Issue #164, opened 2019-08-26
spacing and structure definition for helper-text


SpikedKira:
on https://www.carbondesignsystem.com/components/text-input/style/

spacing between label and input seems to be defined in relation to the label itself in the way of a bottom margin of $spacing-03.

if this is the case, then the spacing between the helper-text and the input is not defined.

the picture however shows a 8px margin on the top of the input.

perhaps margin-top $spacing-03 for the input is simply missing from the table?

aagonzales:
Helper text has recently been update but this issue made me realized we forgot to update the website with the new content. We will work on getting this updated asap!

PR #1421, opened 2020-07-22
docs(style): text input

aagonzales:
Closes https://github.com/carbon-design-system/carbon-website/issues/172 
Related https://github.com/carbon-design-system/carbon-website/issues/164

Also added size variants to the style structure section.


aagonzales:
Oh good catch @laurenmrice. Its updated now.

PR #1433, opened 2020-07-24
pictogram size update

jeanservaas:
Pictogram guidance was wrong here, it allowed for 32px as a minimum size. Minimum size for a pictogram should be 48px and now 112px is allowed... Carbon guidance was updated to match IDL.

- [x] fixed image
- [x] fixed table sizes
- [x] updated paragraph above


conradennis:
I just submitted a PR covering the same updates: https://github.com/carbon-design-system/carbon-website/pull/1438

I noticed the mobile image isn't updated here so maybe we close this in favor of mine?

PR #1461, opened 2020-07-29
feat(project): update to workspace and add gatsby-plugin-feature-flags

joshblack:
This PR introduces a `gatsby-plugin-feature-flags` plugin to work with feature flags in the project. The package itself has a README detailing usage, and is configurable through `gatsby-config.js`. The initial feature flag implemented is `enable-component-index-page` will is used for the component index page as a toggle for displaying it on the site, or not. We are implementing this approach to avoid long-running branches for pages that might take a couple of cycles before they're ready for prod.

#### Changelog

**New**

- Add component index page behind a flag
- Add `gatsby-plugin-feature-flags` plugin

**Changed**

- Shadow nav items from the gatsby theme so that we can conditionally display pages behind a feature flag in the side nav
- Update gatsby config to use new feature flags plugin

**Removed**


vpicone:
@joshblack what if we just didn't include the page in the side bar? We could add an inline notification suggesting the page is a WIP, but it wouldn't be accessible unless someone shared the link.

If we _really_ don't want it to be accessible via prod, we can set an environmental variable in our deploy workflow that results in us ignoing specific directories.

src/gatsby-theme-carbon/util/NavItems.js
```diff
@@ -0,0 +1,42 @@
...
+            return node.name === flag;
+          });
+          return edge.node.value;
+        }
+        return true;
+      }),
+    };
+  });
+
+  return navItems;
```
joshblack, line 41:
```suggestion
```

joshblack:
@vpicone I think the two big things we wanted to avoid were long running feature branches and having stuff we didn't want to show up in prod be visible. Feature flags felt like a nice middle-ground where we could actively develop against `master` and create deployments for demo purposes that would behave like a prod deployment. I definitely think we could replicate this with a single ENV variable, but felt like it was a good point to have a system to help us with any experiments the team wants to run.

vpicone:
@joshblack that makes sense! Do you think maybe we should start with that and add this plugin this once there’s a few flags *and* we need to toggle those flags independently?

Could you elaborate on the concern with just developing off master if we don’t surface the page through the sidebar? Someone would need to be given a the url to find the content. Give directory and adding an inline notification as well could make this really simple. 

Your master branch comment made me think we also have an “experimental” branch w/ a GitHub action to automatically pull any updates from master. It would have the advantage if it’s own vercel url for user testing as well.

joshblack:
> Could you elaborate on the concern with just developing off master if we don’t surface the page through the sidebar?

I think the hope would be to have an environment that would mirror what it would be like in prod. If we were demoing it in a playback it'd be great to have it show up in the sidebar and we can show how to navigate to it from the home page. I think, on the same note, it'd be nice to show the page as-is and not have to worry about displaying banners about it being WIP.

vpicone:
@joshblack I think developing off master (without the link sidenav link) is as close to prod as you can get, that way "turning it on" is as simple as adding the link.

Generating a demo for playback with a sidebar link/notification removed would only take a couple minutes, I'm not sure adding the complexity of a local plugin/feature-flag system is worth it for those moments.

As an example, the theme has a [spacing audit](https://gatsby-theme-carbon.now.sh/test/spacing-audit/) page we use to test vertical rhythm changes and relationships between components. It's nice to have access to it in prod and we haven't had to explain or hide it explicitly.

joshblack:
The proposed approach satisfies the use-cases outlined above and achieves the effect we're looking for in production and in staging sites. I think it's a fair experiment to try out, and it can easily be removed if we find out that it doesn't really work or don't have a need for it after index work is finished 👍 

Let me know if you have any questions or feedback about the PR itself, would love any thoughts you have on it to see if  any improvements could be made 👀

joshblack:
@vpicone definitely see what you're saying with developing off of one page (like the spacing audit) or using an ENV variable to hide pages. I think it's a great option and can definitely go down that route if people prefer it 👍

PR #1465, opened 2020-07-29
feat(data): add index data

joshblack:
Adds in the initial structure for the index data inside of `src/data/index`. This structure should make it easy to manually edit component data for maintainers, as well as make it simple to add new files for each maintainer.

Some of the unsolved problems we have revolve around images at the moment, but this PR does show how the plugin sources the data and is able to be queried by a temporary component index page that we have available.

#### Changelog

**New**

- Added a gatsby plugin for sourcing the index data
  - This plugin also contains some temporary files for converting the CSV from notion to the appropriate YML files in the codebase
  - These temporary files will sit around until the data collection is completed, then should be removed
- Added the community index page with sample data generated from this new data source 🎉 

**Changed**

- Updated the gatsby config to include the plugin for community index data

**Removed**


joshblack:
A couple of notes for this:

- If we group components by "maintainer" then we probably don't need that field anymore on the component YML 👀 
- There should probably be a "validation" check for the YML files to make sure the fields match what's expected

joshblack:
cc @aagonzales @jillianhowarth would love any feedback on the [specific component YML files](https://github.com/carbon-design-system/carbon-website/pull/1465/files#diff-e8a5de03dd26a725d1e5fe7f0261f162) or the [overall directory structure](https://github.com/carbon-design-system/carbon-website/tree/98f22eba9db93dbcbf13377fb167b42a3a4bf63a/src/data/index)!

I'd love to get this in so that @andreancardona can start building off it if you all think the format works 👀 

Let me know if it'd be helpful to add context for the format and contribution cycle too, or could talk about other options if you all had additional ideas I should explore!

aagonzales:
I'm noticing that they're not consistent. Some of the components are missing a category or two sometimes. Does that happen when the category in n/a?

joshblack:
@aagonzales yup, exactly.

aagonzales:
It seems good to me then. I like the folder structure broken down by maintainer. Would you like an official approve on this PR or is a thumbs up here ok? (confusing sense its a draft)

joshblack:
@aagonzales thanks for taking a look! Just wanted to double-check with you all first before adding in checks and what-not for the data 😄 

Will make sure to tag ya when it's ready to go 🚀

joshblack:
FYI @aagonzales @jillianhowarth should be good to go structure-wise, I think the only big change from before is that the maintainers field is no longer required in the YML files since we can get that data from the folder 😄 

There is also a `src/data/index/maintainers.yml` file for listing out the maintainers and their friendly names to be displayed in the UI.

A temporary page should be available at `/community/component-index` to preview the data dump of the ~75ish components that got collected 👀

PR #1496, opened 2020-08-04
chore(project): update index data with latest batch

joshblack:
Update index data with latest batch. This includes several notable changes to the schema:

- `link` has been removed and now we use `website_url` and `code_url`
- Frameworks has been extended to include vanilla
- Availability has been extended to include open source
- `date_added` has been added as a field

#### Changelog

**New**

- Add migration script for new data dump and update it with new structure

**Changed**

- Index data files
- Schema for component files

**Removed**


joshblack:
cc @aagonzales @jillianhowarth this should include the latest updates from the index 👀

Issue #1563, opened 2020-08-12
Add [master dataviz sketch file]: Data viz getting started and resources pages


jeanservaas:
Add a direct download for the WIP master dataviz file so everyone can access it as a resource.

PR #1565, opened 2020-08-13
add[resource card]:master dataviz file direct download

jeanservaas:
Closes https://github.com/carbon-design-system/carbon-website/issues/1563

- [x] Updated Chart anatomy images (desktop and mobile) with new zoom bar and tooltip
- [x] Added master data vis design file direct download resource card to data viz > getting started page
- [x] Added master data vis design file direct download resource card to resources page
- [x] Fixed incorrect donut chart guidance on legend page
- [x] Updated some palettes on legend page that didn't correspond with our docs


theiliad:
🎉  🎉  You did it! this is great

src/pages/data-visualization/chart-anatomy/index.mdx
```diff
@@ -86,17 +86,17 @@ Circular charts are primarily pie and donut charts. Other circular layout charts
...
+    5. Graph frame<br/>
+    6. Big number<br/>
   </Aside>
 </Column>
 </Row>
 
 #### Labels
 
-In a circular chart, labels offer the percentage value of the whole for an individual category.
+In a circular chart, labels offer the percentage value of the whole for an individual category. When the graphic translation of the data is less than 3 degrees, a callout is used to clearly associate the label with the slice. If the data translates as less than 1 degree, a slice will not be rendered on the chart—the data would not be keyboard accessible or available in a tooltip either. The only way to see the data in this scenario would be via a data table, a feature that we'd like to implement for all charts to enhance accessibility.
```
theiliad, line 99:
```suggestion
In a circular chart, labels offer the percentage value of the whole for an individual category. When the graphic translation of the data is less than 3 degrees, a callout is used to clearly associate the label with the slice. If the data translates as less than 1 degree, a slice will not be rendered on the chart, the data would not be keyboard accessible or available in a tooltip either. The only way to see the data in this scenario would be via a data table, a feature that we'd like to implement for all charts to enhance accessibility.
```

theiliad: is that `—` correctly in there?

jeanservaas:
@aagonzales what do you think. If we leave it in box I should add (IBM Internal link) like the other one.

aagonzales:
Yeah I think adding that label would suffice for now.

theiliad:
> @aagonzales what do you think. If we leave it in box I should add (IBM Internal link) like the other one.

does Box do version control? git might be better

theiliad:
What’s stopping us from going live on github right now? Is there a process?
Let’s put this in the hands of everyone rather than just an internal link

PR #1570, opened 2020-08-13
feat(index): add images and update entries

joshblack:
This PR adds in the initial batch of images and sets up the integration for images inside of the component index.

It also addresses some issues I ran into while implementing, namely:

- There was a delay when searching when rendering the whole list that we ran into
- There were some prop type warnings that appeared in console

#### Changelog

**New**

- Add support for images in component index

**Changed**

- Update schema for the metadata of icons to include images
- Update how we render each tile to conditionally render entries
- Update index list to include images
- Update how we do search from an effect to doing it in render

**Removed**


joshblack:
@aagonzales would be great if you have a second to see how the images are rendering (I think they're 4x3 at the moment) to catch any potential issues 👀

Issue #1604, opened 2020-08-24
Heading-03 type styles needs to be updated 


tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/6064

Noticed while working on https://github.com/carbon-design-system/carbon/pull/6714 

Need to update the typestyle table to match changes in code.

PR #1605, opened 2020-08-24
fix(type): update heading-03 line-height

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1604

#### Changelog

**Changed**

- Updates `heading-03` line-height values  to what exists in code

#### Reviewing

@jeanservaas if you could just verify that the rest of the line-height values are correct for the table, I have another PR up using the values directly from the website (https://github.com/carbon-design-system/carbon/issues/6711)


jeanservaas:
Hey @joshblack not sure what I'm doing wrong... I'm trying to change all the scale(5) type line height in the .json file. It should be 24px / 30px line height. But in the preview link it doesn't look like it changed at all. 

And when I inspect, it still says 125% for the line height, which is what is says on the site currently.

image

Issue #1608, opened 2020-08-25
Data Table - isSortable not rendering properly


tw15egan:
Noticed by https://github.com/carbon-design-system/carbon/issues/6715

![Screen Shot 2020-08-25 at 9 12 51 AM](https://user-images.githubusercontent.com/11928039/91200003-72561500-e6b3-11ea-9988-79dec2b650de.png)

PR #1609, opened 2020-08-25
fix(data-table): fix isSortable examples

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1608

Data Table --> Code --> check that the Data Table renders correctly when `isSortable` is checked


emyarod:
related https://github.com/carbon-design-system/carbon-website/pull/1464

PR #1638, opened 2020-08-28
docs: curly quote fix for left nav

janchild:
Adding curly quote to left nav for What’s happening section, for @mjabbink.


mjabbink:
Thank you!

PR #1649, opened 2020-09-01
Tweak to file uploader

chloepoulterdesign:
Closes https://github.ibm.com/CDAI-design/pal/issues/303

We have a CDAI File uploader component which predates the carbon one and is pretty unnecessary now. The only unique guidance it did have was to avoid multiple-file upload in modals in order to avoid scrolling. I'd like to add this to the carbon documentation so that the CDAI component documentation can be deprecated.

On slack agreement from @janchild I've raised this to add that sentence in.

#### Changelog

**New**

- Sentence in the "anatomy" section about modals

**Changed**

- Spotted that the image in "Alignment" said `left align` and the text said `right align` so I've updated it to match the image.


janchild:
@laurenmrice Adding you to this review as you recently updated this documentation. Thanks!

laurenmrice (commented):
As Chloe and I discussed in a call earlier, we will move the sentence about modal guidance into a "When not to use" section.

tw15egan (approved):

chloepoulterdesign:
Thanks @laurenmrice.
I've finally got around to making that change, and it's here:
image

janchild (approved):
🥂  LGTM!!

PR #1660, opened 2020-09-03
Adding dotcom components to index and reordered filters

### LINKED ISSUE #1642: Component Index: Add the "Digital" components to the index
We should make available all the components here - https://www.ibm.com/standards/web/ibm-dotcom-library/components - on the Carbon community component index here - https://www.carbondesignsystem.com/community/component-index.

This should also add "Digital" or "Dotcom" as a maintainer.

shinytoyrobots:
This commit

- Adds "IBM.com" to list of maintainers
- Adds all of the components listed in production at https://www.ibm.com/standards/web/ibm-dotcom-library/ to the community index (_NB: Currently using only default thumbnail - they will need imagery_)
- Reorders the community index filter so that "Maintainer" is the first filter choice

_Note: ComponentIndexPage/index.js fails linting, though not for any of the lines of code I edited, so I commited with --no-very.

Will close #1642


joshblack (approved):

mjabbink:
@jeanservaas @wonilsuhibm We need to get the thumbnails sorted at one point. I think we can publish as is and follow up with those images later. What do you both think?

mjabbink:
@shinytoyrobots The Dotcom components do not show up when checking the box.

Also the AI Applications site we aim to from resource card is this link: https://pages.github.ibm.com/ai-applications/design/ which is different then the site link from the index which is this https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/

I think the new AI Apps site is meant to replace the Watson IoT site but not sure what is going on with the two sites?

shinytoyrobots:
@andreancardona - Can you check with Craig Moser on the correct "AI Applications" site, looks like they didn't only change their name but also changed their primary url?

PR #1666, opened 2020-09-04
Update insiders meeting description

joshblack:
Update insiders description in the events page. This event is currently tied to PALs and framework teams and isn't meant for a general audience.

#### Changelog

**New**

**Changed**

- Update meetups page

**Removed**


shinytoyrobots:
I think the correct resolution here is to write a much clearer description, rather than remove the section entirely.

Similar problem for the Design pin ups (specifically aimed at product teams within IBM, not "anyone who wants some time with the Carbon design team").

So we can be open about the fact that there is an insiders group, and also very specific that it is for stakeholders who are heavy Carbon users to the extent of extending and modifying Carbon specific to their business unit/product space implementation.

emyarod (approved):

joshblack:
@shinytoyrobots no sweat, just updated

janchild (approved):
Looks great, @joshblack!

tw15egan (approved):

PR #1732, opened 2020-09-14
Docs: design tutorials page

jillianhowarth:
Closes #1561 

New page in the Designing section for our Patterns presentation content.

PR #1736, opened 2020-09-15
Docs: 1px grid update

jillianhowarth:
Updating the condensed grid guidance to be 1px to match what's on the new design tutorials page (#1732).

### DIFF
--- a/src/pages/guidelines/2x-grid/implementation.mdx
+++ b/src/pages/guidelines/2x-grid/implementation.mdx
@@ -554,7 +554,7 @@ containers and certain components to “hang” into the gutter.
 <Row>
 <Column colLg={2} colMd={2}>
 
-**Condensed**<br/> 2px gutter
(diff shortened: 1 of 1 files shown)


jeanservaas (commented):
@jillianhowarth looks great! Right after this paragraph under condensed grid section, could you add this paragraph?

The Condensed grid collapses the gutter to 1px. To achieve this, containers and components bleed 16px into the gutter on either side of the column. This grid mode is most often used in conjunction with another grid mode and arguably requires the most design skill to incorporate into layouts.

> In order to emphasize the gutters, 1px borders that are darker than the background color (in both light and dark themes), must be added to any tiles or cards. The borders add just enough clarity to the gutters to stand out, while still registering to the userr as the background color. The token for these borders is $ui-03.

jeanservaas (approved):

tw15egan (approved):

aledavila (approved):

Issue #1744, opened 2020-09-15
For screen reader users component index item tags are buttons with no functionality with additional usability concerns


dakahn:
The component index item tags denoting framework type, design asset type, and maintainer are using the TooltipIcon component which is a `<div role="button">`. This is problematic for both mouse and screen reader users. 

![image](https://user-images.githubusercontent.com/40970507/93274347-c66a8b80-f77f-11ea-8ec3-d3634d03f0a0.png)
Mouse users see a pointer cursor prompting them to interact with the element, but it does nothing since the information is revealed on hover and there is no other functionality. More generally, it seems strange to hide this information in a tooltip -- since the framework type and asset type isn't additive information, but is instead potentially crucial to selection. 

In my screen reader testing with NVDA on Windows 10 the three tags are described to the user as "buttons", 
> list  with 3 items  Vanilla  Vanilla  button    Vanilla
Sketch  Sketch  button    Sketch
Maintainer  Watson Health  button    Maintainer

This is misleading since they have no functionality (the tooltips are visual affordances only -- the screen reader user relies on the elements programmatic naming). 

We have a unordered list of three elements -- two icon buttons and a "tag" or pill with two different ways of conveying information. 
![image](https://user-images.githubusercontent.com/40970507/93275060-7c82a500-f781-11ea-918b-e469359e191b.png)
The icon buttons let the mouse user discern framework or asset type through iconography OR carefully hovering over a 24px by 24px target area and _perhaps_ discovering that by clicking they can cause the tooltip to persist (a long walk for a magnification user to take), while the third pill lists the name of an internal team hoping the user hovers to find out that this team is the component's maintainer. 

Both screen reader and mouse users could benefit from simplifying things a little here. Spelling out clearly `Vanilla JS`, `Sketch` and `Maintainer: Watson Health` would definitely be more accessible and potentially more usable. _(this is my naïve solution without fully knowing design/implementation constraints)_

tw15egan:
Since it's already being read as a button, I wonder how hard it would be to have it filter by that type when you click on it. So say a user clicks the `React` tag, it would update the index to only show React packages

PR #1745, opened 2020-09-15
fix(index item tags): hide icons from a11y tree; add descriptive aria labels

dakahn:
ref #1744 

In my screen reader testing the index tags were being read twice -- once for the list item itself and another time for the image (icon) inside of it. It was also difficult to discern the information the tags were trying to convey. This PR does two things:

- hides the icons from the accessibility tree with `aria-hidden`
- adds more descriptive aria labeling so the information (framework type, design asset type, and maintainer) is more understandable
- BONUS THING: removes the word 'image' in our asset placeholder which was being flagged by jsx-a11y as redundant and unnecessary given it's context in the accessibility tree (cool!)

### note for testing
The elements are still read as buttons which is confusing. I started a more thorough discussion of this at the referenced issue above 🏄

### DIFF
--- a/src/components/ComponentIndexPage/ComponentIndexListItem.js
+++ b/src/components/ComponentIndexPage/ComponentIndexListItem.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/display-name */
 /**
  * Copyright IBM Corp. 2016, 2020
  *
@@ -61,10 +62,7 @@ const ComponentIndexListItem = React.memo(
       );
     } else {
       img = (
-        <img
-          src={placeholder}
-          alt={`Placeholder image for the ${name} component`}
-        />
+        <img src={placeholder} alt={`Placeholder for the ${name} component`} />
       );
     }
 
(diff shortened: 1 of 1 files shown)


src/components/ComponentIndexPage/ComponentIndexListItem.js
```diff
@@ -61,10 +62,7 @@ const ComponentIndexListItem = React.memo(
       );
     } else {
       img = (
-        <img
-          src={placeholder}
-          alt={`Placeholder image for the ${name} component`}
-        />
+        <img src={placeholder} alt={`Placeholder for the ${name} component`} />
```
emyarod, line 65: just wondering is this formatting change intentional or was it based on personal settings?

dakahn: I should just be picking up _our_ prettier config, unless something changed with VS Code/prettier

tw15egan (approved):
LGTM! 👍 ✅

src/components/ComponentIndexPage/ComponentIndexListItem.js
emyarod: ok I see, strange that it was not formatted before for whatever reason

emyarod (approved):

Issue #1647, opened 2020-08-31
Replicate the community index model for data visualization/charts components


shinytoyrobots:
Following a similar pattern, we should build an index for data visualization. This should showcase all existing charting components, with maintainer info etc, but also allow submission of other charts and under different frameworks.

So should model, very similarly, to https://www.carbondesignsystem.com/community/component-index.

Should be able to submit entries in a similar way (via .yml file)

Should include thumbnail, title, description, link to website guidance for that chart, link to code (where available).

Filter categories should be the same (even if actual filter items might be different - e.g. different maintainer list).

_NB: If end point is possible integration of this index and the component index, in a single "catalog" of resources, then database scheme should be consistent/compatible.

@joshblack, @andreancardona - anything in terms of pitfalls that we should be aware of?
@aagonzales - could we use the same design template for this (at least for initial pass)

shinytoyrobots:
@aledavila - let me know if you've had a chance to look into this yet? We can schedule some pairing time next week when I'm back in office.

PR #1746, opened 2020-09-16
repurposing component index code for charts index

shinytoyrobots:
Addressing #1647 

At this point primarily duplicated the community component index into a "Charts" component, and relevant GraphQL schema.

Very much WIP.

Need to refine from a code / functionality perspective.

Need design feedback on possible changes to layout for a chart index vs a component index.


shinytoyrobots:
If it builds this time (!), also interested in a sanity check from @joshblack / @andreancardona. At this point it really is primarily a near-complete duplicate of all the component index code and schema. 

I think that should work as a foundation for any other type of index (as we'll end up doing likewise for patterns at some point)...although also feels like maybe longer term there's a better way than just rebuilding three different indices with three different (but basically identical) schemas. :D

joshblack:
Definitely could try and repurpose the index plugin too to work for both types, would be great if we could do:

```js
{
  resolve: 'gatsby-plugin-index',
  options: {
    sources: [
      {
        name: 'community',
        directory: '...',
      },
      {
        name: 'charts',
        directory: '...',
      },
    ],
  },
}
```

And it would just work. In the future might even make sense for this data just to be coming from an API and editable through a UI (which could be do-able with our service that we're doing right now since we'll have component data there)

shinytoyrobots:
That makes sense.

Presumably then we could just add all the chart .yml files in the interim, even if they're initially being referenced by `chart-index`, and then change to reference them from a combined index in future (i.e. we add the data first).

This is, I think, definitely the correct approach. Because then we'll end up with (as a final result)...

```js
{
  resolve: 'gatsby-plugin-index',
  options: {
    sources: [
      {
        name: 'components',
        directory: '...',
      },
      {
        name: 'charts',
        directory: '...',
      },
      {
        name: 'patterns',
        directory: '...',
      },
    ],
  },
}
```

...where we can present the different types on different pages (so a "component index", "chart index" etc), but *also* in theory a single, filterable, "catalog of everything".

shinytoyrobots:
@andreancardona / @joshblack - this works ok, but it breaks completely if you filter to a point of "no results". Do you have any thoughts?

src/components/ChartIndexPage/ChartIndexNotFound.js
```diff
@@ -0,0 +1,44 @@
...
+        </Column>
+      </Row>
+      <Row>
+        <Column sm={4} md={6} lg={6}>
+          <p className="component-index-not-found__text">
+            This community component index does not include the Carbon core
+            components and you may find what you're looking for in the
+            <Link
+              className="component-index-not-found__link"
+              href={coreComponentsLink}
```
joshblack, line 32:
```suggestion
              href={coreChartsLink}
```

joshblack:
@shinytoyrobots took a quick look, I _think_ the fix is that suggestion above 👀  let me know if it works!

src/components/ChartIndexPage/ChartIndexNotFound.js
shinytoyrobots: Yep, that was it! Thanks.

shinytoyrobots:
Now that this has some population with data - I think it's important that we get some design feedback. @aagonzales, I'll set up a brief chat with you, @aledavila and I about any considerations we should take into account. Let me know if there's anyone else I need to add to that meeting.

shinytoyrobots:
@andrea-gm or @joshblack...

If I were to update the `schema.js` for this so that "Chart Type" was an array;

`chart_type: Joi.array().items(Joi.string().valid('Comparisons', 'Trends', 'Part-to-whole', 'Correlations', 'Connections', 'Geospatial: overlays', 'Other')),`

...how would I update the ChartIndexPage component's search function so that it returns data correctly for the filter.

If I just update the schema to make Chart Type array rather than string, the "Chart Type" filter as currently written just returns no results (as it's presumably searching for a single string and not an array item).

shinytoyrobots:
@jeanservaas / @aagonzales - from a data/functionality perspective this is pretty much ready now. What would we want to do in terms of thumbnail images (I made a couple of pngs using the svgs on the carbon site to test how they looked)?

shinytoyrobots:
Moving this out of draft (though it still needs visuals), as it is now functional.

shinytoyrobots:
I added updated images for all of the Carbon Charts types (with the exception of "step chart", which doesn't have a thumbnail).

jeanservaas:
@janchild are we cool with calling it a "charts" index or are we leading with the name "data visualization" — we've been inconsistent on this front. 

@shinytoyrobots I will reformat the thumbnails like I did for the others so the margins are roughly the same, but that doesn't need to block this from going live.

janchild:
@jeanservaas I'd prefer we go with Data visualization if that fits. It's a mouthful but I think when people say it, they'll shorten it to data viz index anyway.

shinytoyrobots:
OK, let me update with that language - see how it looks. And also just do a minor bit of cleanup (remove a couple of AI Apps solutions that don't have descriptions right now, and remove the Cognos test chart!)

aagonzales (commented):
Just a minor thing that doesn't need to hold up release either but at the small/medium breakpoint the maintainer tags drop and I think there's plenty of space for them to stay here. 

image

shinytoyrobots:
That's a good point @aagonzales. Presumably also applies on the component index.

Especially in the example you show, where there are duplicate entries, maintainer info to distinguish between the two seems important.

shinytoyrobots:
There's a minor bugginess in the "Previous/Next" at the bottom of the page (same on the component index), which isn't a stop shipper but flagging here so we don't forget.

joshblack:
@shinytoyrobots I think it would need to be an update to this method:

https://github.com/carbon-design-system/carbon-website/pull/1746/files#diff-817efa7c15d29e19212dd673e7ce745bR82

Specifically:

```diff
function filterItems(items, filters) {
  return items.filter((item) => {
    // If there are no filters, return true.
    if (filters.length === 0) {
      return true;
    }
    const { chartType, availability, maintainer, complexity } = item;
-    const fields = [chartType, availability, maintainer, complexity];
+    const fields = [...chartType, availability, maintainer, complexity];
    return filters.every((filter) => fields.includes(filter));
  });
}
```

This _should_ add every item from chartType to the "fields" array that is checked against for each filter 👀

andreancardona:
@shinytoyrobots @aledavila For the "no chart found" results do we want to add a sentence or two linking them to `contribute a chart`?

Screen Shot 2020-06-18 at 2 56 26 PM

shinytoyrobots:
> @shinytoyrobots @aledavila For the "no chart found" results do we want to add a sentence or two linking them to `contribute a chart`?
> 
> Screen Shot 2020-10-07 at 3 03 07 PM

Good thinking, @andreancardona. I added that to last push.

@joshblack - thanks for the feedback. I'll test that as a post initial release thing, I think.

@janchild / @mjabbink - are we good with the "Data visualization index" language now?

If all that's OK then we can approve and merge I think.

janchild:
@shinytoyrobots yes, the language looks good. We might take a look at the labeling of the tabs to remove the repetition, but this is good for now. 

Tricky because I don't think tabs can have the same name, is that correct @alisonjoseph ?

andreancardona (approved):

mjabbink:
I think that intro copy needs to speak to the designs expressing IDL and quality. I’d remove anything that isn’t. 

Most of the ones in AI Apps are just referencing existing carbon charts and the rest are barely anything more than a gesture.

A few also are misnamed in the index like bumble chart leads to bullet in AI Apps.

mjabbink:
Looking at the AI Apps charts I think these need to probably not be included in the index. Only 5–6 new ones and they are incredibly weak with no guidance, specs, or refined designs, Sketch files zero. Just a static image with missing details and misleading. 

We have current design efforts on following charts:

**Bullet
KPI
Network Diagram
Histogram
Alluvial 
Lollipop**

shinytoyrobots:
Part of the point of the index is to highlight duplicates though - to provide a greater impetus to consolidate and refine. So would we be better keeping the AI Apps charts in (for a soft launch at least) and using that as a basis to reach out to Craig to discuss effective next steps?

alisonjoseph:
Is there a reason we are duplicating the code from the component index and not reusing the same component for both? Would we want to just have one component for both so its easier to update in the future and keep them in sync?

aledavila:
@alisonjoseph so that will actually be a next step. To repurpose the component plug-in to allow for us to reuse it and be able to add the charts and patterns.

vpicone:
@shinytoyrobots you can't have two properties on an object with the same name. This PR broke all the component index images.


Screen Shot 2020-10-08 at 9 51 21 PM


Screen Shot 2020-10-08 at 9 53 17 PM

shinytoyrobots:
Thanks for the spot. And a good reminder that just testing the thing being built can mean missing knock-on effects (from clumsy coding on my part).

PR #1767, opened 2020-09-17
quick add some images for dotcom comps

shinytoyrobots:
Added some of the thumbnails from other maintainer groups, that roughly matched the dotcom maintained community components(so they don't all have the same default thumbnail).


shinytoyrobots:
From discussion with @jeanservaas yesterday, dotcom have their own component images. So we can hold off on moving forward with this PR until we incorporate those images (rather than an interim solution).

shinytoyrobots:
Updated the 13 components we had from dotcom in the index with the correct images - those can be merged once approved.

nb. Used existing `card.png` used from other PALs; a) for consistency and b) dotcom didn't have a specific thumbnail.

Noticed that dotcom also added a number of components - will need to add those in a separate PR.

aagonzales (approved):

joshblack (approved):

PR #1832, opened 2020-09-24
Watson-processing

heatherco:
Various modals to inform users of Watson processing states.

Watson evaluating - screen 1
Watson evaluating - screen 2
Watson evaluating - screen 3
Evaluation complete

Watson evaluating - Error during processing
Watson evaluating - Watson is almost done - error during processing

### DIFF
--- a/src/data/index/watson-processing
+++ b/src/data/index/watson-processing
@@ -1,0 +1,14 @@
+name: component name
+website_url: >-
+  url of documentation / guidance page (if any)
+code_url: >-
+  url of component code on GitHub
+maintainer: name of maintainer team
+description: >-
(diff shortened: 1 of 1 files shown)


joshblack:
Hey there, @heatherco! 👋 

It seems like you're trying to add an item to the index. When looking at the YML file, it seems like it hasn't been filled out so it won't show up correctly in the index. Could you take a moment and fill it out and then we can review the changes? 👀

tw15egan:
Hey @heatherco, looks like there hasn't been any action taken on these PRs, so we're going to close them out for now. If you'd like these items to be added to the index, please fill out the `.yml` file and we can reopen the PR. Thanks!

PR #1839, opened 2020-09-24
updated naming for cdai to cloud + cognitive

shinytoyrobots:
Updating references to "CD+AI" to reference "Cloud & Cognitive", following their updated naming.


janchild (approved):

joshblack (approved):
Looks good, quick question: do we need to change the directory name too? (from cdai)

Issue #1831, opened 2020-09-24
FormGroup component in Form doc missing required prop 


queeniema:
## Summary

Issue here: https://www.carbondesignsystem.com/components/form/usage#live-demo

The code snippet provided is the following:

```
<Form>
  <FormGroup>
    <TextInput
      helperText="Optional helper text here; if message is more than one line text should wrap (~100 character count maximum)"
      id="test2"
      invalidText="Invalid error message."
      labelText="Text input label"
      placeholder="Placeholder text"
    />
  </FormGroup>
...
```

When opening the example in CodeSandbox, I see this error in the Console:

```
Warning: Failed prop type: The prop `legendText` is marked as required in `FormGroup`, but its value is `undefined`.
```

According to the Storybook, `legendText` is in fact required, but the sample code does not include that prop.

<img width="1019" alt="Screen Shot 2020-09-24 at 8 35 19 AM" src="https://user-images.githubusercontent.com/2454818/94167138-00ebbc80-fe41-11ea-8a40-ee9048bb9a52.png">

queeniema:
@tw15egan I see that the `FormGroup` components were removed and I don't see any other references to them on that doc page. In what scenarios should they be used?

tw15egan:
@queeniema Seems like they are mainly used to wrap `RadioButtonGroup` and `Checkbox` so that they have an associated `legend` element to work as a label.

Issue #1828, opened 2020-09-23
Sticky color theme switcher positioning


mattrosno:
It's sitting low when you scroll.

![image](https://user-images.githubusercontent.com/1691245/94072109-b32c7100-fdba-11ea-98b7-89a651932c49.png)

mjabbink:
![ezgif com-optimize (1)](https://user-images.githubusercontent.com/32881239/94200720-7c209300-fe80-11ea-9679-a8eea199df09.gif)

aagonzales:
Duplicate issue, I also believe it's already fixed. 
https://github.com/carbon-design-system/carbon-website/issues/1302

PR #1841, opened 2020-09-24
fix(website): fix FormGroup warnings, color table bugs

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1831
Closes https://github.com/carbon-design-system/carbon-website/issues/1828

Our `FormGroup` component requires a `legendText` prop, which is used to populate a label for groups like Radio Buttons and Checkboxes. All other form elements supply their own label through the `labelText` property.

Also, fixed an issue with the color table hanging over the table on scroll. 

#### Changelog

**Changed**

- swapped `FormGroup` with generic `div` and added spacing 
- Modified some Layout styles for when the banner is/is not present. 

**Testing** 
- [ ] Ensure `Form` and `TextInput` Usage and Code pages look correct
- [ ] Ensure all sticky headers (Color Table, Typography Table, etc..) render properly with/without the banner


joshblack (approved):
Changes look good 👍  for the website is this a common enough change where we want a class name instead of the inline styles for the new `<div>`s? Just wasn't sure when looking at the diff 👀

tw15egan:
I can create a class for this. I added some notes in our Prop audit around this convention, and how we need to normalize form spacing 👍 . Seems like we were using`FromGroup` just to get that nice `2rem` of space

tw15egan:
@joshblack actually on second thought, since these are copyable code snippets it may be confusing to add a class that will have no styles associated with it when copied over.

aledavila (approved):

aagonzales (approved):
Looks good!

PR #1857, opened 2020-10-01
fix(TypesetStyle): add missing 'id' to resolve a11y issue

### LINKED ISSUE #1510: audit-avt1-guidelines-typography: invalid id
## Detailed Description:
The value "screenWidthInput" of the 'for' attribute is not the 'id' of a valid <input> element

Screen Shot 2020-08-04 at 1 42 44 PM

## What to do:
Set the value of the for attribute to the id value of the valid <input> element to be labeled. Make sure the id value is unique. For example:

## Why is this important?
A label element is attached to a specific <input> element through the use of the for attribute. When a <label> does not reference the corresponding unique id in a valid lableable <input> element, assistive technologies will not be able to associate the label with the element.

kleyu:
Closes #1510 

add missing `id` for input

#### Changelog

**New**

- Added `id` prop to `InputRange`

### DIFF
--- a/src/components/TypesetStyle/InputRange.js
+++ b/src/components/TypesetStyle/InputRange.js
@@ -4,8 +4,9 @@ import { settings } from 'carbon-components';
 
 const { prefix } = settings;
 
-const InputRange = ({ step, min, max, value, onChange }) => (
+const InputRange = ({ step, min, max, value, onChange, id }) => (
   <input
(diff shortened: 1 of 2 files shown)


emyarod (approved):
looks good to me

tw15egan (approved):
Thanks for contributing! 👍 ✅

PR #1860, opened 2020-10-01
fix: make img alt attribute empty

### LINKED ISSUE #1819: audit-avt1-data-viz-types: link text repeated
## Environment:

> Operating system: Mac
> Browser: Chrome 
> AVT :one:
> Page: [Data visualization - Chart types](https://www.carbondesignsystem.com/data-visualization/chart-types)

## Detailed Description: 

Link text is repeated in an image 'alt' value within the same link

Screen Shot 2020-09-22 at 4 17 46 PM

## Why is this important? 
When the text equivalent of an image inside or adjacent to a link is the same as the link text, people with visual impairments who use a screen reader will hear the same text announced twice.

## What to do: 
If a link has text and a redundant image, set the image alt attribute to alt=“” (null alt without any spaces);

motou:
Closes #1819 

set the image alt attribute empty so that screen reader won't annonce the text twice.

#### Changelog

**Changed**

- change the image alt text to empty string
 See: https://www.gatsbyjs.com/plugins/gatsby-remark-images/?=gatsby-remark#usage-in-markdown


tw15egan (approved):

kleyu:
correct me if I'm wrong, but I think `` syntax erases `alt` tag altogether and this is not ideal, the desired behaviour is to have `alt=""` [1] according to decorative images spec.

IMO this is some constraint on `gatsby-remark-images` plugin as they don't allow for `null`  alt text as far as I know.


[1] https://www.w3.org/WAI/tutorials/images/decorative/

tw15egan:
@kleyu nice catch! You are correct, seems like this is currently not possible (passing in an `alt=""`) due to Gatsby setting a default value if one is not passed. Looks like we'll need to get this fixed upstream before we can fix it on our end.
Issue: https://github.com/gatsbyjs/gatsby/issues/20179
Gatsby source: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js#L173

motou:
> @kleyu nice catch! You are correct, seems like this is currently not possible (passing in an `alt=""`) due to Gatsby setting a default value if one is not passed. Looks like we'll need to get this fixed upstream before we can fix it on our end.
> 
> Issue: [gatsbyjs/gatsby#20179](https://github.com/gatsbyjs/gatsby/issues/20179)
> Gatsby source: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js#L173

Okay, then let me fix that issue at first 🙃

motou:
@tw15egan  thanks for approving. The fix PR https://github.com/gatsbyjs/gatsby/pull/27218 is already merged. When the new version 3.3.33 is published, then this PR is ready to be merged as well.

pieh:
> IMO this is some constraint on `gatsby-remark-images` plugin as they don't allow for `null` alt text as far as I know.

(I'm maintainer of `gatsby-remark-images`). The behaviour like that (falling back to filename as alt) was unfortunate choice in the past, that I hope to remove in the next major release (probably will have a way to maintain that behaviour for anyone that wish to keep it). There is also opportunity to add global plugin setting to disable this behaviour in current major version line too (without causing breaking changes and forcing major version bump just yet)

tw15egan:
@pieh @motou thanks for working together so quickly to tackle this! 🙏 🥇

motou:
@tw15egan since the new version 3.3.33 is already published (https://github.com/gatsbyjs/gatsby/pull/27218#issuecomment-704155858), this PR is ready to be merged :)

andreancardona (approved):

andreancardona:
@motou yay! 🥳  thanks for the contribution!

Issue #1875, opened 2020-10-07
Footer on site not showing any version info for React


shinytoyrobots:
![image](https://user-images.githubusercontent.com/825457/95354262-d3176680-0889-11eb-9c87-8c76e1e0d114.png)

Should be showing latest version info rather than blank space.

PR #1878, opened 2020-10-07
fix(footer): ensure correct react version is shown

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1875


emyarod (approved):
looks good to me

joshblack (approved):

PR #1881, opened 2020-10-08
Config tweaks

vpicone:
1. Updates theme (patch)
2. Update fathom usage to conform with their [new script/SPA recommendations](https://usefathom.com/integrations/gatsbyjs)
3. Remove unnecessary fathom/netlify plugins


gatsby-config.js
```diff
@@ -45,12 +45,12 @@ module.exports = {
       },
     },
     {
-      resolve: 'gatsby-plugin-fathom',
+      resolve: 'gatsby-plugin-manifest',
```
emyarod, line 48: does this plugin need to be in package.json as well?

vpicone: It's shipped with the theme but it's a good idea to add it explicitly, I'll add it.

PR #1883, opened 2020-10-08
fix(component-index): fix horizontal scrolling on component index page

### LINKED ISSUE #1628: Community assets —> Component index: Page should be locked in mobile
In mobile (Chrome) the screen moves all around. Should be locked for left and right horizontal movement.

@andreancardona

aledavila:
Closes #1628 

Fixes the jumping around of the index page on mobile. The issue came from the tags in the component view adding some random extra space. 

#### Changelog

**Changed**

- changed the tags in the footers to absolute positioning.

Check mobile component index page


andreancardona (approved):
yay!

src/components/ChartIndexPage/chart-index.scss
```diff
@@ -1,223 +0,0 @@
-//
```
joshblack, line 1: Is this file being used by the Chart Index work that Robin was working on?? 🤔

jeanservaas (approved):

src/components/ChartIndexPage/chart-index.scss
aledavila: No I think we can remove it with no issue. It was only copied over. But I noticed that all the styles were the same so I removed them and nothing broke.

aledavila:
I removed the file and it doesnt seem to mess anythign up. Mind giving it one more look :) @joshblack

alisonjoseph (approved):

PR #1900, opened 2020-10-16
Added the full width grid modifier

wannymiarelli:
The grid documentation actually tells you to look in the "full-width" example to learn more about the "full-width modifier" for the grid. But there is no such example in the page.
I think it would be a great idea to add this class directly in the paragraph.

#### Changelog

**Changed**

- the "container" paragraph in the Grid documentation page https://www.carbondesignsystem.com/guidelines/2x-grid/implementation/ added reference to the full width grid modifier class.


joshblack (approved):

jillianhowarth (approved):

andreancardona (approved):

andreancardona:
@wannymiarelli looks good!

PR #1922, opened 2020-10-26
Remove data viz index

alisonjoseph:
Request from @mjabbink & @jeanservaas to remove the data viz index from the site. 

> Ultimately, it’s not really an appropriate format for the content. Data viz is about libraries, not individual charts. Also, Chart types (e.g. bar chart) are a lot more like patterns than components in that there are many varieties of each type. 

*I left all the data files and images for future implementation of an index but can remove if that makes more sense?*


jeanservaas (approved):
Thanks Alison!

janchild (approved):

tw15egan (approved):

Issue #1915, opened 2020-10-23
[live demo] update accordion component to include disabled state


laurenmrice:
- The disabled state was recently added to the accordion component. ([PR](https://github.com/carbon-design-system/carbon/pull/7005))
- The disabled knob needs to show up in the live demo.

<img width="1137" alt="Screen Shot 2020-10-23 at 9 48 38 AM" src="https://user-images.githubusercontent.com/43969356/97011680-fe01eb80-1514-11eb-83af-eaf04a0535d7.png">

PR #1923, opened 2020-10-26
docs(Accordion): update Accordion knobs to include new props

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/1915

Adds in `disabled` props to `Accordion` demo. Also updates packages to the newest versions.

#### Changelog

**New**

- New knobs on the `Accordion` demo

**Changed**

- Updated `react-docgen.json`

**Removed**

- `Accordion` overrides from `overrides.json`

Ensure all component demos still work as expected


tw15egan:
Noticing an issue when disabling a single `AccordionItem` the disabled border styles are being passed to all AccordionItems

tw15egan:
Pushed up a fix here: https://github.com/carbon-design-system/carbon/pull/7142

jillianhowarth (commented):
This is awesome! One note - it doesn't look like there's a way to go back to the default accordion size once another size is selected.

tw15egan:
@jillianhowarth that sort of due to how that prop is used, we only allow `sm` and `xl` as options for size... to get the default size, you just wouldn't pass in the size prop i.e `size="xl"`

jillianhowarth:
@tw15egan right, that makes sense in the code. My concern is that folks using the demo will click `sm` or `xl` and then not realize that there’s a third default size and not be able to get back to it if they want to see that size. It seems like in the demo we should have a third option for `default` the way we do for some of our other components

joshblack (approved):

tw15egan:
We only allow `sm` or `xl` as options

Screen Shot 2020-10-27 at 12 03 28 PM

Just going to remove the size switcher for now, as this behavior can be seen in the react storybook and I don't want to spend too much time duplicating functionality

jillianhowarth (approved):
Sounds good!

aledavila (approved):

PR #1948, opened 2020-11-03
fix: add white background back to images

alisonjoseph:
Updates token so that there should now be a white background behind usage page images

### Testing

Need to make sure this doesn't add a white background to images where it shouldn't

### DIFF
--- a/src/styles/_images.scss
+++ b/src/styles/_images.scss
@@ -17,11 +17,11 @@
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  background: $ui-02;
+  background: $ui-01;
 }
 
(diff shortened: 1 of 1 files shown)


emyarod (commented):
do you have a specific page or image we can check to confirm this is appearing correctly? I'm not sure off the top of my head which images already have backgrounds

alisonjoseph:
Any of the usage pages for components should work.

aagonzales (approved):
🎉 Looks good!

aagonzales:
Its actually most images on the style tab. Date Picker was the component I spotted it first.

emyarod (approved):
looks good to me

johnbister (approved):

PR #1960, opened 2020-11-05
fix: restore function to DataTable live demo knobs

### LINKED ISSUE #1959: DataTable demo knobs not functioning properly
## Detailed description

> Describe in detail the issue you're having.

When adjusting the knobs in the DataTable component live demo, the options are not properly reflected in the demo component.

> What did you expect to happen? What happened instead? What would you like to see changed?

The demo knobs should control the component props and change accordingly

## Steps to reproduce the issue

Visit https://carbondesignsystem.com/components/data-table/usage/ or https://carbondesignsystem.com/components/data-table/code/, modify the live demo knobs, and see that no changes are reflected for several knobs

emyarod:
Closes #1959

When adjusting component live demo knobs, the KnobContainer would concatenate the default knob props after the newly added props, which would result in the new values being overwritten by the defaults. The regular expression to separate the default props from the component tag name would also result in some faulty matches in render prop arrow functions. This PR replaces render functions in the DataTable component demo and switches the order of default props and newly added props in the component demos.

#### Changelog

**Changed**

- place default props before modified props in knob containers
- refactor DataTable render props to render with children instead

**Testing / Reviewing**

Confirm that DataTable demos are functioning as expected, and other component demos remain as is


tw15egan (approved):
Woo! Awesome fix 🎉 👍 ✅

jillianhowarth (approved):
This is great! 🙂

aledavila (approved):

PR #1963, opened 2020-11-06
Update style.mdx

aagonzales:
Updated incorrect color values for UI Shell header.

Oh actually wait this might be wrong, production does show this as the original values. Let me investigate why/when this changed.


jillianhowarth (approved):

aagonzales:
Opps yeah the original docs are correct. The kit and images are wrong. Going to close this.

PR #1968, opened 2020-11-09
chore: remove alert banner

### LINKED ISSUE #1304: [Style regression] Sticky bar not accounting for BLM banner
I believe the BLM banner is messing up where the size toolbars are sticking in the UI 
https://www.carbondesignsystem.com/guidelines/typography/expressive#fluid-type-sets

image

### LINKED ISSUE #1967: Remove voting banner
- Explore turning into a component
- Disable it or remove it entirely

vpicone:
Closes #1967 
Closes #1304

I looked into turning the Alert banner into a component, but I think for the initial implementation I was a bit rushed. We could improve upon it next time using the git history if its helpful.


src/gatsby-theme-carbon/components/Layout.js
```diff
@@ -32,7 +24,6 @@ const Layout = ({
   tabs,
 }) => {
   const is404 = children.key === null;
-  const [isBannerVisible, setBannerVisibility] = useState(true);
```
tw15egan, line 35: Do we perhaps just want to set this to false? That way when we want to reenable the banner we can just set this back to true // a different local storage item

vpicone: @tw15egan I feel a bit better about just removing it all together so we don't have unused ghost styles and code. Like you said, we'll want to revisit our local storage approach next time and that makes up the bulk of the component.

We can always come back and reference this implementation if needed.

andreancardona (approved):
looks good @vpicone !

tw15egan (approved):

PR #1969, opened 2020-11-09
Update eslint config

vpicone:
- Removes all the wesbos stuff
- Still uses Carbon prettier config (via eslint) for formatting


.eslintrc
```diff
@@ -1,14 +1,31 @@
...
     "fathom": "readonly"
   },
+  "env": {
+    "node": true,
+    "browser": true,
+    "commonjs": true,
+    "es6": true
+  },
+  "parser": "babel-eslint",
+  "extends": [
```
joshblack, line 13: Would it be helpful to try out: https://www.npmjs.com/package/eslint-config-carbon ? Might help with some of the base rules and then can layer on the import ones 👀

vpicone: @joshblack for sure, can we update the peer dep to allow eslint 7?

joshblack: Hell yeah 🔥

dakahn (approved):
Seems cool to me -- if we're still doing this and pending whatever other convos are going on

Issue #1925, opened 2020-10-27
[checkbox] update live demo to show correct disabled state


laurenmrice:
The live demo on the checkbox tab currently shows incorrect disabled state styles. It is currently a bug in the demo, but correct in react.

<img width="1148" alt="Screen Shot 2020-10-27 at 12 30 52 PM" src="https://user-images.githubusercontent.com/43969356/97339270-bfdc3300-1858-11eb-89e7-7b9f3a862060.png">

laurenmrice:
This is still happening but now with the checkbox group header.
<img width="1108" alt="Screen Shot 2022-10-25 at 11 39 26 AM" src="https://user-images.githubusercontent.com/43969356/197819005-455417f1-0f06-4e11-810b-da8316d3ade1.png">

alisonjoseph:
Replaced with storybook demo https://github.com/carbon-design-system/carbon-website/pull/3716

PR #1977, opened 2020-11-12
fix(KnobContainer): update KnobContainer to account for self-closing elements

tw15egan:
Rel https://github.com/carbon-design-system/carbon-website/pull/1960
Rel https://github.com/carbon-design-system/carbon-website/issues/1925

Closing slash was being placed before the added knob props, causing the component to render incorrectly. This PR changes it so that it will place the slash at the end of the newly added props on self-closing elements.

#### Changelog

**New**

- `KnobContainer.js` checks for trailing slash in `defaultKnobProps`, and if it is true removes the slash before concating with the newly added knob props. Then adds in the closing slash at the end of the component string. 

Open to ideas on how to solve the linked issue, the disabled states for checkbox are not tied to the element but rather to the wrapper and needs to be added in manually or use the `FormGroup` component


tw15egan:
@emyarod just tagged you since I know you just modified this as well 👍

joshblack (approved):

emyarod (approved):
looks good to me

PR #1988, opened 2020-11-17
chore: update deployments to point toward main branch

jnm2377:
I created a branch called `main` with all the commit history that `master` has BUT before switching over to main as our default branch, I need to update where our deployment points to:

#### Changelog

- updates deployments to point toward `main` instead of `master`
- _Note: this PR is being compared to/will be merged into the `main` branch, so it shouldn't affect anything on `master` or the current deployed website until we switch default branch over to main_

Lmk if there's any other config I missed :)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
     inputs:
       name:
-        description: "Reason"
+        description: 'Reason'
   push:
     branches:
-      - master
+      - main
 
 jobs:
   build:
@@ -35,7 +35,9 @@ jobs:
       - name: Login to ibmcloud
         env:
           API_KEY: ${{ secrets.API_KEY }}
-        run: ibmcloud login -a "https://cloud.ibm.com" -u apikey -p "$API_KEY" -o "carbon-design-system" -s "production" -r "us-south"
+        run:
+          ibmcloud login -a "https://cloud.ibm.com" -u apikey -p "$API_KEY" -o
+          "carbon-design-system" -s "production" -r "us-south"
 
       - name: Deploy website
         run: |
--- a/.github/workflows/process-images.yml
+++ b/.github/workflows/process-images.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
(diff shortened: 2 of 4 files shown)


.github/workflows/process-images.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@master
+        uses: actions/checkout@main
       - name: Compress Images
-        uses: calibreapp/image-actions@master
+        uses: calibreapp/image-actions@main
```
joshblack, line 16: Does calibreapp use a `main` branch? Just wasn't sure 👀

.github/workflows/deploy.yml
```diff
@@ -35,7 +35,9 @@ jobs:
       - name: Login to ibmcloud
         env:
           API_KEY: ${{ secrets.API_KEY }}
-        run: ibmcloud login -a "https://cloud.ibm.com" -u apikey -p "$API_KEY" -o "carbon-design-system" -s "production" -r "us-south"
+        run:
+          ibmcloud login -a "https://cloud.ibm.com" -u apikey -p "$API_KEY" -o
```
joshblack, line 39: If prettier is formatting this line this way, we may need to use the `|` character for multiline just to make sure the bash command works as expected 👀

joshblack (commented):
Looking good! Just left a couple of comments/questions 👀

.github/workflows/process-images.yml
jnm2377: oh that's a good question lol I guess I should check! I suppose maybe not, since we've been using master for the config and that hasn't broken lololol 😅

.github/workflows/deploy.yml
jnm2377: @joshblack should I just include that character at the end of the first line? I thought the multiline character was `\`? then again, bash is not my strong-suit so I may be making things up lol

.github/workflows/process-images.yml
jnm2377: yeah, they're on master. I'll update this. Good catch!

tw15egan (approved):

joshblack (approved):

PR #1992, opened 2020-11-18
Import Pattern Accessibility documentation

chloepoulterdesign:
There isn't an issue for this, however I am acting on the output of Fang Shih's audit of pattern accessibility. You can see her findings [here](https://ibm.box.com/s/m7bh9pul6p57fgedjse5193rxyn66ujp) (w3 log-in required).

#### Changelog

**New**

- Section called accessibility
- Anchor link 
- Modal & Single file sub-section stepping through triggering options and keyboard interaction orders
- Full page & side panel sub-section covering keyboard interaction orders


tw15egan (approved):

jeanservaas (changes requested):
Hi @chloepoulterdesign thanks!

Some of what I commented on the export pattern will have the same impact here:

image

I'm going to attach those templates that I built for full screen UI here too so that we're all using the same aspect ratios, margins (depending on the aspect ratio) and background browser asset.

Please make sure these images are using either a 16:9 or a 3:2 aspect ratio... a little taller image might be nice because the bottom / top margin is a little tight. I think that add file button can also be attached to the input field in these cases.

image

Would you get that native OS red number for dragged files in product? Also, I know Mac on desktop has the red number but the round tags around the file names do not look like native mac.

image

We don't use drawn arrows in the design language. I think you can make your point about tabbing with text only, either above the image or below in the caption

In dropdown, we talk about key commands and show similar sequences, there might be some good inspiration here:

https://www.carbondesignsystem.com/components/dropdown/usage

image

In this section, I might actually keep the side panels examples and guidance together under one sub head and the full page guidance together under another subhead so the user's not jumping around 

image

Also, I think some of the 608 usage full screen templates will work for you here and give you more margin / eliminate that shadow.

image

Same with Export, I'm going to give you the anatomy callout symbols that we use in a sketch file. Also I might restructure the content a little here. Anatomy images are usually only used once or twice in a pattern. A lot of the parts that you see in the initial image are also in the following images. What you're showing here seems to be more of a sequence of behaviors rather than an anatomy. 

I think we actually need better guidance around a situation like this. I'm not saying you can't do this, but I think you could edit this area down a lot so there's less redundancy. 

In Carbon we tend to isolate pieces like this to show their states (file uploader is a good usage tab to look at for what you've got going:

image


image

Also, keep all margins even in situations like this, and try not to use a margin of less than 32px all the way around

image

Remember your feedback and related sections too. File uploader for instance would be related and the Create flow too.

chloepoulterdesign:
On discussion with @jeanservaas I've removed the imagery in the accessibility section. 
I've tidied things up to make it clearer without images. 
I've renamed the sections to better explain that the behaviour varies depending on whether it is a `single` or `multi` file import/upload. 
I've raised #2013 to work on the image quality in the pattern documentation in future.

jeanservaas (commented):
Hey @chloepoulterdesign 

Same apology here, you're not responsible for updating images that are already on the site, haha sorry. 

As for the added accessibility content, could we treat it the same way as you did for the Export pattern added content, which I just approved (i.e. following the Carbon pattern and not trying to add all those anatomy images)?

chloepoulterdesign:
Hey @jeanservaas
I have made the same changes as in Export here - it looks like the Vercel deployment failed when I committed them.
I can't trigger the redeployment, could you or someone in the team have a go at that?
image

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+- On `tab`, apply focus to the "remove" icon on the added file(s), moving
+  through each using `tab`.
+
+#### On reaching the maximum number of added files
+
+In some usecases, there are a maximum number of allowed files. In this situation
+the file selection methods are in disabled state. Focus cycles between the
+selected files "remove" actions, and moving to the next section of the page.
+
+1. On returning to the page/panel from the final File uploader/"Add file"
```
jillianhowarth, line 281:
```suggestion
1. On returning to the page/panel from the final file uploader/add file
```

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+   interaction, (after any optional loading state) focus is applied to the
+   "remove" icon on the uploaded files.
```
jillianhowarth, line 283:
```suggestion
   remove icon on the uploaded files.
```

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+2. On keying `tab`, focus moves through the "remove" icon of each uploaded file.
```
jillianhowarth, line 284:
```suggestion
2. On keying `Tab`, focus moves through the remove icon of each uploaded file.
```

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+3. On continued use of `tab`, focus is applied to the next section or
```
jillianhowarth, line 285:
```suggestion
3. On continued use of `Tab`, focus is applied to the next section or
```

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+   interaction in the page.
+
+## Related
```
jillianhowarth, line 288:
```suggestion
```

jillianhowarth (changes requested):
@chloepoulterdesign this is looking great! Just some minor capitalization/punctuation fixes 🙂

src/pages/community/patterns/import-pattern/index.mdx
```diff
@@ -175,7 +175,118 @@ If your product supports importing from a URL, use this method.
...
+#### Trigger
+
+Often, the upload or import modal (page, or panel) is launched by a button
+press.
+
+1. The user navigates to the Import button using the tab key. If Import is
+   launched from an overflow menu, the user may navigate to Import within the
+   overflow menu using the up/down arrow keys.
+2. The focus state is applied to the button/menu item.
+3. The user clicks “enter” to launch the Import modal.
```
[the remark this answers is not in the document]
chloepoulterdesign: Should this be `Return` rather than `Enter`?

jeanservaas (approved):
lgtm thanks Chloe!!

jillianhowarth (approved):
Great work!! 🙂

PR #2000, opened 2020-11-20
DO NOT MERGE chore(release): smoke test v10.25.0-rc.1

aledavila:
smoke test for release v10.25.0-rc.1


dakahn (approved):

andreancardona (approved):

joshblack (changes requested):
Looking great so far, one thing for pictograms is that I think we'll need to remove the rule for setting the stroke to `currentColor` as it is causing the pictograms to render each path thicker than intended

Issue #1936, opened 2020-10-29
fix[data viz color palette]: light and dark theme palettes are not switching


jeanservaas:
In the pairings switcher, the when users toggle between Light and Dark tabs, the palettes do not change.

The component is displaying the dark theme palette under the Light tab as well as the Dark tab, whereas it should display the Light theme palette when the user selects the Light tab:

![image](https://user-images.githubusercontent.com/43144260/97634991-a1b33600-1a04-11eb-8d72-d7b183a0de1b.png)

![image](https://user-images.githubusercontent.com/43144260/97635066-bdb6d780-1a04-11eb-92d1-756c59b0560f.png)

The same error is happening with the Alert palette switcher, only the light theme palette is showing for both the Light and Dark tabs :

![image](https://user-images.githubusercontent.com/43144260/97635289-15edd980-1a05-11eb-9e64-cceeaf9749b2.png)

![image](https://user-images.githubusercontent.com/43144260/97635294-1c7c5100-1a05-11eb-9d13-1fd543b37233.png)

mjabbink:
@alisonjoseph This needs to be prioritized big time. Not understanding why this is not resolved in 25 days. @jeanservaas and @alisonjoseph please keep me posted on issue. Needs to be closed next week please.

tw15egan:
@mjabbink pr is up to fix the grouped color palettes, but it looks like there is only 1 palette for `alerts` across both themes

https://github.com/carbon-design-system/carbon-website/blob/79071bc7ef6d6fe236d987738af3c59b4d57514c/src%2Fdata%2Fdata-visualization%2Fpalettes.js#L1171-L1192

@jeanservaas is this incorrect?

PR #2007, opened 2020-11-24
fix(data-viz): ensure color palette updates on theme change

tw15egan:
https://github.com/carbon-design-system/carbon-website/issues/1936

Fixes an issue where the color palette was not updated when the theme was changed

#### Changelog

**Changed**

- `useEffect` hook to update the color values when the theme is changed, not just when the dropdown is changed

### DIFF
--- a/src/components/ColorPalette/ColorPalette.js
+++ b/src/components/ColorPalette/ColorPalette.js
@@ -16,7 +16,8 @@ import {
   fiveColorDark,
   monoColors,
   divergingColors,
-  alert,
+  alertLight,
+  alertDark,
 } from '../../data/data-visualization/palettes';
 import ColorPaletteColor from './ColorPaletteColor';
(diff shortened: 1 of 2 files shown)


jeanservaas (commented):
thanks for getting on this! Just a couple of things:
That dropdown (for the color groupings) to the right of the tabs should also have a white (not g10 background)


image

The alert palette also changes from light to dark, right now those are only the light values on both of the tabs and green value is wrong for the light:

image

image


**Should be:**
image

tw15egan:
@jeanservaas updated, although I noticed `#198038` (rather than `#78a9ff`) seems like the correct hex for `Green 60`.

emyarod (approved):
looks good to me pending visual review

aledavila (approved):

jeanservaas (approved):
boom

PR #2025, opened 2020-12-02
chore(release): smoketest for v10.25.0-rc2

aledavila:
smoke test for v10.25.0-rc.2

Correct versions for following are:
React: v7.25.0-rc.3
Pictograms: v11.0.0-rc.3

Testing:

Make sure pictograms is correct


joshblack:
This will have the same issue as before with the styles for pictograms, correct? 👀 

Feel free to keep using the same PR btw for RCs, especially since this one has some style tweaks that need to happen. Then when you publish to latest you can add in the latest dependencies

andreancardona:
> e same issue as before with the styles for pictogram

@joshblack  yup - just pushed up the change for the styles scaling.

joshblack (approved):

tw15egan:
Is the increase in size intentional? 

Screen Shot 2020-12-02 at 1 16 22 PM

andreancardona:
> Is the increase in size intentional?
> 
> Screen Shot 2020-12-02 at 1 16 22 PM

no I committed a change to bring it back down to scale to match what we currently have on the site - looks like it's still deploying. @tw15egan

tw15egan (approved):

PR #2035, opened 2020-12-07
docs(content): update type to variant

aagonzales:
We've decided to go with the term "variant" instead of "type" to describe that different kinds of variations within a component. Updated all instances or "type" to read "variant" instead.


src/pages/components/breadcrumb/usage.mdx
```diff
@@ -50,9 +50,9 @@ If you are taking users through a multistep process use a
 
 ### Types
 
-Carbon supports two types of breadcrumbs. Both types are styled the same, but
+Carbon supports two types of breadcrumbs. Both tyoes are styled the same, but
```
janchild, line 53:
```suggestion
Carbon supports two types of breadcrumbs. Both types are styled the same, but
```

src/pages/components/breadcrumb/usage.mdx
```diff
@@ -50,9 +50,9 @@ If you are taking users through a multistep process use a
...
 the methods for populating the breadcrumb trail are different. The breadcrumb
-type used should be consistent across a product.
+variant used should be consistent across a product.
```
janchild, line 55:
```suggestion
type used should be consistent across a product.
```

janchild (approved):
Awesome, @aagonzales! Just two small fixes but then it's good to go.

joshblack (approved):
Looks great!

PR #2055, opened 2020-12-18
docs: new pattern templates

janchild:
This PR updates the existing single pattern template with two types of pattern templates--one for single variant, and one for multiple variants. 

The pattern templates themselves have undergone significant revision to capture what we learned through both the pattern development process and the component usage guidance development process.

### DIFF
--- a/src/pages/contributing/component/index.mdx
+++ b/src/pages/contributing/component/index.mdx
@@ -13,39 +13,22 @@ enhancements or brand new assets.
 </PageDescription>
 
 <AnchorLinks>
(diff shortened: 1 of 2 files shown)


alisonjoseph (approved):
Looks good. Might want to add line breaks between the anchor links in the code snippet.

jeanservaas (approved):

janchild:
> Looks good. Might want to add line breaks between the anchor links in the code snippet.

Thanks, @alisonjoseph. I did this by adding two spaces at the end of each line. It did the trick!

PR #2128, opened 2021-02-05
docs: button variant guidance refinements

janchild:
This PR is to add clarity around the use of button variants. In particular, the correct use of a secondary button as the negative option for a primary button, and being clear that it cannot be used in isolation or for a positive action. 

This came out of discussions in the January 21. Design Crit.

### DIFF
--- a/src/pages/components/button/usage.mdx
+++ b/src/pages/components/button/usage.mdx
@@ -76,7 +76,7 @@ actions.
 | _Secondary_ | For secondary actions on each page. Secondary buttons can only be used  in conjunction with a primary button. As part of a pair, the secondary button's function is to perform the negative action of the set, such as Cancel or Back. Do not use a secondary button in isolation and do not use a button for a positive action. |
 | _Tertiary_  | For less prominent, and sometimes independent, actions. Tertiary buttons can be used in isolation  or paired with a primary button when there are multiple calls to action. Tertiary buttons can also be used for sub-tasks on a page where a primary button for the main and final action is present.                           |
 | _Danger_    | For actions that could have destructive effects on the user’s data  (for example, delete or remove). Danger button has three styles: primary, tertiary, and ghost.                                                                                                                                                               |
-| _Ghost_     | For the least pronounced actions. Ghost buttons are often used in conjunction with a primary button, or with a primary and secondary button pair.                                                                                                                                                                                |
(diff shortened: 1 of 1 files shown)


src/pages/components/button/usage.mdx
```diff
@@ -65,13 +70,13 @@ desired action is to take the user to a new page.
...
+| _Tertiary_  | For less prominent, and sometimes independent, actions. Tertiary buttons can be used in isolation 
or paired with a primary button when there are multiple calls to action. Tertiary buttons can also be used for sub-tasks on a page where a primary button for the main and final action is present.                           |
+| _Danger_    | For actions that could have destructive effects on the user’s data 
(for example, delete or remove). Danger button has three styles: primary, tertiary, and ghost.                                                                                                                                                               |
+| _Ghost_     | For the least pronounced actions. Ghost buttons are often used in conjunction with a primary button, or with a primary and secondary button pair.                                                                                                                                                                                |
```
aagonzales, line 79: > or with a primary and secondary button pair

I think this point might need more clarification. I have Friday brain so I can't think of how to clarify it but maybe even just saying in "rare scenarios it can be paired ... ." I think those scenarios mostly happen in progress flow and the ghost button tends to be the "cancel"

aagonzales:
I wonder if we should add a section specifically about "button pairing"

janchild:
> I wonder if we should add a section specifically about "button pairing"

Yes, I thought about that but I was trying to keep it all up top with the basic guidance where people would see it. But maybe we could have preliminary copy here and then a section that goes into more detail with examples.

emyarod (approved):
looks good to me

src/pages/components/button/usage.mdx
janchild: @aagonzales just want to make sure I've got this right. With the progress flow example, the secondary button would still be a negative action, like Back, and the ghost would be a complete opt-out like Cancel. Is that right?

aagonzales: Yup, exactly

janchild:
> > I wonder if we should add a section specifically about "button pairing"
> 
> Yes, I thought about that but I was trying to keep it all up top with the basic guidance where people would see it. But maybe we could have preliminary copy here and then a section that goes into more detail with examples.

Created a new issue for a more in-depth treatment and some image review --> https://github.com/carbon-design-system/carbon-website/issues/2142

janchild:
@aagonzales this should be good to go, if you want to take a peek at the latest!

aagonzales (approved):

Issue #1498, opened 2020-08-04
audit-avt1-guidelines-icons: form button needs label


andreancardona:
## Environment:
> Operating system: Mac
> Browser: Chrome
> AVT :one:
> Page: [Guidelines - Icons: Library](https://www.carbondesignsystem.com/guidelines/icons/library)
> This bug affects about 95 other issues 

## Detailed Description:
Form control element <button> has no associated label

<img width="1680" alt="Screen Shot 2020-08-04 at 12 18 49 PM" src="https://user-images.githubusercontent.com/32720851/89319294-e883d080-d64d-11ea-8a54-9cadbccb774e.png">

## What to do:
Add a for attribute to a visible label with the id of this control (e.g. <label for="controlID">);

## Why is this important?
Associating a meaningful label with every UI control allows the browser and assistive technology to expose and announce the control to a user.

andreancardona:
Fixed in monorepo: https://github.com/carbon-design-system/carbon/pull/7812

andreancardona:
Fixed in PR: https://github.com/carbon-design-system/carbon/pull/7814

PR #2144, opened 2021-02-17
fix(icons): use icon button instead of definition tooltip

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/7821
Refs https://github.com/carbon-design-system/carbon-website/issues/1498

The use of `TooltipDefinition` as icon-only buttons to download/copy SVG assets in the icon library was throwing a11y violations. This changes them to use the icon-only ghost `Button` component instead, which resolves these violations. This should resolve violations on the Pictograms page as well

#### Changelog

**Changed**

- Uses `Button` with icon-only props instead of `TooltipDefinition` 

**Testing**

Ensure the Icon library pages still look the same and all functionality remains


andreancardona (approved):
Tested and approved - thanks for doing this! :)

jnm2377 (approved):

jnm2377:
cc: @vpicone @aledavila not sure if you are aware of this PR/issue, but would be good to update in the IDL iconography page as well if it has the same issue.

vpicone:
@tw15egan do you mind pulling these changes into the IDL site as well?

PR #2157, opened 2021-02-26
fix(breadcrumb): demo matches usage guidelines

### LINKED ISSUE #2125: Breadcrumbs live demo is misleading
The [live demo for the Breadcrumbs](https://www.carbondesignsystem.com/components/breadcrumb/usage) shows the current page in the breadcrumb and a trailing slash by default. The default, according to the documentation, is for the current page to **_not_** be in the breadcrumb. And when the current page **_is_** shown in the breadcrumb, there should **_not_** be a trailing slash.

The ask for the live demo and the Breadcrumb implementation (to avoid developer confusion and inconsistencies):

1. don't show the current page by default + do show the trailing slash by default
2. allow adding the current page as an option + do not show the trailing slash when this option is selected

jnm2377:
Closes #2125 

**Changed**

- Breadcrumb demo shows current page demo without trailing slash

**Reviewers**

- Test out live demo in the `usage` and `code` tabs for Breadcrumb component

### DIFF
--- a/src/pages/components/breadcrumb/code.mdx
+++ b/src/pages/components/breadcrumb/code.mdx
@@ -52,16 +52,31 @@ usage documentation, see the Storybooks for each framework below.
 
 <ComponentDemo
   components={[
+    { id: 'breadcrumb', label: 'Breadcrumb' },
     {
-      id: 'breadcrumb',
-      label: 'Breadcrumb',
+      id: 'currentPage',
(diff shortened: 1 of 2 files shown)


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

jillianhowarth (approved):

emyarod (approved):

PR #2167, opened 2021-03-08
[button] update icon-only guidance

laurenmrice:
We do not support a danger icon-only button. It only is available in the primary, secondary, tertiary and ghost variants.

**Changed**

- Icon-only buttons: Edited out that we support danger icon-only buttons.

### DIFF
--- a/src/pages/components/button/usage.mdx
+++ b/src/pages/components/button/usage.mdx
@@ -632,9 +632,7 @@ a text label, but use icon buttons cautiously.
 
 ### Icon-only buttons
 
-Icon buttons allow users to take actions, and make choices, with a single tap.
-Icon buttons can take the form of any of the five variants above but most
-commonly will be styled as primary or ghost buttons.
+Icon buttons allow users to take actions, and make choices, with a single tap. Icon buttons can take the form of a primary, secondary, tertiary or ghost variant but most commonly will be styled as primary or ghost buttons.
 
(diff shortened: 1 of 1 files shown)


src/pages/components/button/usage.mdx
```diff
@@ -632,9 +632,7 @@ a text label, but use icon buttons cautiously.
 
 ### Icon-only buttons
 
-Icon buttons allow users to take actions, and make choices, with a single tap.
-Icon buttons can take the form of any of the five variants above but most
-commonly will be styled as primary or ghost buttons.
+Icon buttons allow users to take actions, and make choices, with a single tap. Icon buttons can take the form of a primary, secondary, tertiary or ghost variant but most commonly will be styled as primary or ghost buttons.
```
janchild, line 635:
```suggestion
Icon buttons allow users to take actions, and make choices, with a single tap. Icon buttons can take the form of a primary, secondary, tertiary, or ghost variant but most commonly will be styled as primary or ghost buttons.
```

janchild (approved):
@laurenmrice This is great! Just one comma tweak. ;)

emyarod (approved):
looks good to me after suggested fixes

src/pages/components/button/usage.mdx
laurenmrice: You got me !!!

PR #2180, opened 2021-03-15
chore: add dco action

mattrosno:
Adds GitHub Action to enforce Developer Certificate of Origin (DCO) signatures. Read more here: https://github.com/carbon-design-system/carbon-dco

Configuration:

1. I added the `carbon-bot` PERSONAL_ACCESS_TOKEN to this repository as a secret
2. I made sure to exclude dependabot and kodiakhq


.github/workflows/dco.yml
```diff
@@ -0,0 +1,30 @@
...
+        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the DCO document and I hereby sign the DCO.') || github.event_name == 'pull_request_target'
+        uses: cla-assistant/github-action@6cfe025afc31eb0bdd08e23d3f2e9ed141ab2c09
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: "dco-signatures.json"
+          path-to-document: "https://github.com/carbon-design-system/carbon-dco/blob/main/dco.md"
+          branch: "main"
+          allowlist: dependabot,kodiakhq
```
tw15egan, line 22: Will Vercel also trigger this when it comments the deploy previews?

What about PR's by another action like https://github.com/carbon-design-system/carbon-website/pull/2177

tw15egan (approved):
A small comment, but verified it works on the test repo 👍🏻

.github/workflows/dco.yml
mattrosno: `kodiakhq` bot makes the commits for the image compression PRs, so I think we're good there. E.g. https://github.com/carbon-design-system/carbon-website/pull/2177/commits/2b48d3b5a44e16d169c49ce0135ce4757594398e

I'm unsure if this check runs on every comment like Vercel deploy previews. If the action is slowing down our review, we can always disable it and look for alternative solutions.

tw15egan: Yeah, just something I was thinking of, definitely not a blocker. Thanks for adding this in!

tay1orjones (approved):

joshblack (approved):

mattrosno:
I'm going to update this to use a newer version of the GitHub Action, then we can merge it in.

mattrosno:
Testing the check. It takes a minute or so before the check runs.
image

Signing the DCO.
image

It took a few minutes for the DCO to automatically run again after signing (commenting).
image

The check passes.
image

`carbon-bot` added this commit in the `carbon-dco` repo to save the signature. https://github.com/carbon-design-system/carbon-dco/commit/f39e6cc5da7cadb3eb5fc6273e46cf3318925c44

mattrosno:
@tw15egan let's wait on this issue that kodiak's primary maintainer just opened. https://github.com/chdsbd/kodiak/issues/641

I can also do a test with kodiak to ensure if this is just an issue with this specific PR because it adds the GH action that references the repo secret (meaning we can just merge this PR without the check), or if it'll be an ongoing issue with future PRs after this has been merged.

PR #2187, opened 2021-03-17
fix(overlay): remove overlay on mobile

tw15egan:
Rel https://github.com/carbon-design-system/carbon/pull/7891

Prevents overlay from showing on mobile by default, unless it is expanded


jnm2377 (approved):
Working as expected for me! 🔥

tay1orjones (approved):

PR #2261, opened 2021-04-29
fix: carbon-overrides path in Angular Tutorial

igaldino:
In Carbon Angular Tutorial, step 2, it was pointing to a wrong path for carbon-overrides.scss file.


janchild (approved):

emyarod (approved):
looks good to me

tw15egan (approved):

PR #2291, opened 2021-05-07
Fixes angular tutorial

igaldino:
A few fixes to Angular tutorial steps 3 and 5.


janchild (approved):
LGTM!

src/pages/developing/angular-tutorial/step-5.mdx
```diff
@@ -152,7 +152,7 @@ first.
 npm run build
 ```
 
-Looking at `package.json`, you'll find `yarn build` to run `ng build`. This
+Looking at `package.json`, you'll find `ng build`. This
```
cal-smith, line 155: Just curious why the switch away from `yarn build`? `npm run build` also seems a better choice

src/pages/developing/angular-tutorial/step-5.mdx
```diff
@@ -262,10 +262,10 @@ sure we only deploy immediately after running the build. In the `"scripts"`
 object in `package.json`, add:
 
 ```bash path=package.json
-"deploy": "rm -rf ./dist && yarn build && cf push -f manifest.yml"
+"deploy": "rm -rf ./dist && ng build --prod --aot && cf push -f manifest.yml"
 ```
 
-Next time you want to deploy, you can simply run `yarn deploy`.
+Next time you want to deploy, you can simply run `npm run-script deploy`.
```
cal-smith, line 268: Same question as above, I believe we use `yarn` everywhere else, so the change is strange. If we want to go with `npm` instead this should be `npm run deploy`

cal-smith (changes requested):
Thanks for taking the time to update this! Just a couple questions, but it seems good otherwise!

src/pages/developing/angular-tutorial/step-5.mdx
igaldino: Sure, in the tutorial, this is the only mention to yarn.

src/pages/developing/angular-tutorial/step-5.mdx
igaldino: In the tutorial, this is the second mention to yarn.

About `npm run deploy`, I didn't try but `npm run-script deploy` works just fine too.

igaldino: the rest of angular tutorial uses ng or npm

cal-smith: Ah, my mistake on the `yarn` situation ... This seems fine then

tw15egan (approved):

PR #2303, opened 2021-05-13
fix tooltip style override issues in charting demos

### LINKED ISSUE #2302: All charting tooltips are broken - style override
**Demo:**
https://www.carbondesignsystem.com/data-visualization/simple-charts/#bar-(floating)

image

I believe this is because of this override
image

theiliad:
Closes #2302

### DIFF
--- a/src/components/ComponentDemo/ComponentDemo.module.scss
+++ b/src/components/ComponentDemo/ComponentDemo.module.scss
@@ -439,7 +439,7 @@ body[data-live-preview-theme='Gray 100'] ul[class^='bx--overflow-menu-options'],
 }
 
 body {
-  [class$='--tooltip'] {
+  [class$='bx--tooltip'] {
     background-color: var(--cds-inverse-02, #393939);
   }
 


theiliad:
@emyarod could you pls review this PR? It overrides a change you pushed last week and I want to make sure that your intended functionality also isn't broken by the change in my PR.

Also, are we able to add `bx` to the rest of your added styling overrides?

emyarod:
off the top of my head I think this should be fine, but I will need to check if any of the other tooltip variants are affected by this

theiliad:
> off the top of my head I think this should be fine, but I will need to check if any of the other tooltip variants are affected by this

ok, lmk pls, and if we can, let's add `bx` to the rest of the styles that you added last week, hopefully in a separate PR

emyarod:
the other styles were following the precedent set in the existing selectors, which AFAIK is to avoid hard coding the prefix. would it be safer to just ignore `--cc--tooltip` instead of targeting `bx--tooltip`?

theiliad:
> the other styles were following the precedent set in the existing selectors, which AFAIK is to avoid hard coding the prefix. would it be safer to just ignore `--cc--tooltip` instead of targeting `bx--tooltip`? (although I don't mind hard coding the prefix for all of the selectors)

Could be, however that doesn't future proof the work, does it? E.g. I see you have modal in there, charts will soon use the Carbon Components' modal component, which could again cause some collisions here. Haven't checked that example yet but just something to keep in mind

emyarod (approved):

emyarod:
yeah it would be good to know which components you are using/will be using, and I think we can increase the specificity of the selectors as needed. it may not be just adding the prefix across the board since it looks like many of the selectors are matching against suffixes

theiliad:
> yeah it would be good to know which components you are using/will be using, and I think we can increase the specificity of the selectors as needed. it may not be just adding the prefix across the board since it looks like many of the selectors are matching against suffixes

I'm wondering why we'd prefer not to use the Carbon prefix when addressing styles? That'd be the safest path IMO, since Carbon Charts also adds its own prefix to styles (e.g. `bx--cc--tooltip`), that way we'll know for sure that we won't collide

emyarod:
I was looking back at the original commits that started using those kinds of attribute selectors and couldn't exactly tell why (#1111), but I assume it had to do with the uncertainty around keeping the `bx` prefix going forward

theiliad:
> I was looking back at the original commits that started using those kinds of attribute selectors and couldn't exactly tell why (#1111), but I assume it had to do with the uncertainty around keeping the `bx` prefix going forward

we could easily import the prefix from Carbon rather than hardcoding it, that way whatever it changes to, we support that.

Nowhere in the Carbon Charts' styles is bx hardcoded, we just do this:

`@import './vendor/@carbon/type/scss/prefix';`

```scss
svg.#{$prefix}--#{$charts-prefix}--chart-svg {
	overflow: visible;
}
```

emyarod:
yeah I was only referencing the website styles which target some attributes with specific suffixes rather than specifying further, like how this current PR is moving from selecting `--tooltip` suffixed attributes to `bx--tooltip` suffixed attributes

tay1orjones (approved):

PR #2324, opened 2021-05-18
feat: add experimental ListSection component from IDL

vpicone:
closes #2323

This PR adds an experimental implementation of what a ListSection component might look like so we can see where it works or is falling short.

**To view the test page, add `/test/list-section/` to the preview URL.**

PR #2336, opened 2021-05-20
chore(gatsby): remove v3.0 `FAST_REFRESH` flag

emyarod:
related #2324

hot reloading appears to be broken based on user reports. It seems that the `FAST_REFRESH` flag is not supported in the version of Gatsby we're currently using (ref https://github.com/gatsbyjs/gatsby/discussions/28390), and removing the flag restores hot reloading

not sure if we're ready to upgrade Gatsby to 3.0+ but in the meantime it seems that we can't use this flag yet

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -12,7 +12,6 @@ module.exports = {
   flags: {
     PRESERVE_WEBPACK_CACHE: true,
     FAST_DEV: true,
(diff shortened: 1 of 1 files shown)


vpicone (approved):

andreancardona (approved):
nice catch!

PR #2354, opened 2021-05-25
 docs: v11 color token updates for file uploader component

laurenmrice:
**Changed**
 - This PR updates the color tokens on the Style tab for the file uploader component.
    - Changed column `Class` --> `Element`
 
---

**New**
- Added Sizes table
- Added Sizes image


joshblack (approved):

aagonzales:
I'm wondering if for the size table if the size column should come first? Or at least second. It feels a little buried for being the title information. 

image


We should probably get consistent with this for when we update size prop names next. This is what I did for text input. It does get tricky when there are two elements together. Maybe it should go Elements | Size | Property | Value. I don't think you need the spacing token category unless we want to change that category to [size token](https://github.com/carbon-design-system/carbon/pull/8671).

image

laurenmrice:
@aagonzales Yes, it does get buried I agree. I reordered the columns for the Size table to be `Elements | Size | Property | Value` for now until we align on the structure of the tables for when we tackle size props. I also took out the Spacing token column. 👍

aagonzales (approved):
Looks good!

PR #2370, opened 2021-06-01
chore(project): remove add-ons contribution docs

joshblack:
This PR removes the add-ons contribution page as it is no longer needed. It's also the beginning of the conversation around how we would like to best document the process of creating/maintaining PALs at IBM.

#### Changelog

**New**

**Changed**

**Removed**

- Removed add-on page as they are not currently being developed at IBM


joshblack:
cc @mattrosno @alisonjoseph @jeanservaas curious what you all think about this!

alisonjoseph (approved):
this makes sense to me. Kill it 👍

PR #2376, opened 2021-06-03
Tab style v11

kingtraceyj:
### Updating Style tab

**Updating:**

- Color tokens
- Interactive states
- Structure

**Adding**

- Size props


emyarod (approved):
looks good to me

aagonzales (commented):
I just noticed this, you're using Plex Mono in your spec images. They should actually be using just Sans now. We've updated that at one point so some of the old images still have mono.

src/pages/components/tabs/style.mdx
```diff
@@ -33,17 +166,73 @@ Tab label should be set in sentence case, and should not exceed three words.
...
+</Column>
+</Row>
+
+<Caption>
+  Structure and spacing measurements for contained tabs | px / rem
+</Caption>
+
+## Sizes
+
+| Element                 | Property | Height (px / rem) |
```
aagonzales, line 222: The second column should say "Size" not "Property"

tw15egan:
The page is pretty long, should we have anchor links at the top of the page? Otherwise this looks great 👍🏻 

Screen Shot 2021-06-03 at 3 27 09 PM

src/pages/components/tabs/style.mdx
```diff
@@ -33,17 +166,73 @@ Tab label should be set in sentence case, and should not exceed three words.
 
 ## Structure
 
-| Class                 | Property                    | px / rem | Spacing token |
+### Line tabs
```
aagonzales, line 169: This is the way I would structure this table 

| Element         | Property                    | px / rem | Spacing token |
| --------------- | --------------------------- | -------- | ------------- |
| Tab             | border-bottom               | 2px      | –             |
|                 | margin-left                 | 1px      | —             |
| Label           | padding-left, padding-right | 16 / 1   | `$spacing-05` |
|                 | padding-top, padding-bottom | 8 / 0.5  | `$spacing-03` |
| Tab icon        | padding-right               | 16 / 1   | `$spacing-05` |
|                 | padding-left                | 8 / 0.5  | `$spacing-03` |
|                 | svg (md)                    | 16 x 16  | —             |
|                 | svg (lg)                    | 20 x 20  | —             |
| Scrollable icon | svg                         | 16 x 16  | –             |

src/pages/components/tabs/style.mdx
```diff
@@ -33,17 +166,73 @@ Tab label should be set in sentence case, and should not exceed three words.
...
+</Caption>
+
+## Sizes
+
+| Element                 | Property | Height (px / rem) |
+| ----------------------- | -------- | ----------------- |
+| Line tab                | Medium   | 40 / 2.5          |
+| Icon-only line tab      | Medium   | 40 / 2.5          |
+|                         | Large    | 48 / 3            |
+| Contained tab           | Large    | 48 / 3            |
```
aagonzales, line 227: I think since contained tabs only has one size (large) I would take it out of the size table and just include that information as `height` its structure section. I'd then make the heading "size" and h4 under the contained tab section.

It then might help to add a sentence about why line tab icon only has a size variation but the normal line tab doesn't. I think this section is more confusing then it needs to be.

aagonzales:
@tw15egan I've thought this same thing about anchor links. I think a lot of our style tabs could use them.

src/pages/components/tabs/style.mdx
aagonzales: I added height and weight as per my other comment. I think then the image would read like 
image

aagonzales: I would also maybe hyperlink "auto width" in the table to where you talk about it on the usage tab.

tw15egan (approved):
LGTM 👍 ✅

src/pages/components/tabs/style.mdx
```diff
@@ -33,17 +174,102 @@ Tab label should be set in sentence case, and should not exceed three words.
...
-| --------------------- | --------------------------- | -------- | ------------- |
-| `.bx--tabs__nav-item` | border-bottom               | 2px      | –             |
-| `.bx--tabs__nav-link` | padding-left, padding-right | 16 / 1   | `$spacing-05` |
-| `.bx--tabs__nav-link` | padding-top, padding-bottom | 8 / 0.5  | `$spacing-03` |
-| `bx--tabs__nav-item`  | margin-left                 | 1px      | -             |
+### Line tabs
+
+| Element         | Property                    | px / rem   | Spacing token |
+| --------------- | --------------------------- | ---------- | ------------- |
+| Tab             | height                      | 40 / 2.5   | –             |
```
aagonzales, line 181: Its hard to tell in markdown but looking at the preview some of your dashes are different length. They all should be using the en dash —

aagonzales (commented):
I DMed you about this but the structure images should be using the fixed image component. Sorry I didnt catch that early.  

And then just two minor content things

src/pages/components/tabs/style.mdx
kingtraceyj: made sure all were en dashes!

aagonzales (approved):
Looks good to me! Ship it!

PR #2377, opened 2021-06-03
feat: add StatusIndicator table component

### LINKED ISSUE #2199: feature[patterns]:modified table for status indicator pattern
## feature[pattern]:modified table for status indicator pattern

We are enhancing and migrating the Status Indicator pattern from the Carbon for Cloud & Cognitive PAL over to the Carbon website. A table like this already exists on that PAL, I've just modified it slightly.

https://pages.github.ibm.com/cdai-design/pal/patterns/status-icons

### Modifications of the C&CC table

- We've removed one of the columns (the status column)
- We added the outlined versions of the status icons
- We increased the width of the g100 strip in the table so that it's on the responsive column grid at all breakpoints
- We are specifying that the rule, when it is on the g100 background pull in the $interactive-02 token (g80), not the $ui-03 token (g20); i.e. the rule will be g20 on the white background and g80 on the dark background, as if the dark area is a dark theme zone.

### Specs and Behavior
The vertical spacing, type hierarchy and horizontal scrolling behavior will operate just like a normal Gatsby table, except this one has a white background. At mobile, the table will scroll horizontally, just like the C&CC table does. 

At the table and mobile breakpoints, the dark logos have been removed so that the table can still fit at 672px and will be more light weight in the mobile experience.

### Full layout (@vpicone the new content is in the file)

Status icons_table_content

### 1312

image

<br>

image

<br>

### 1056

image

<br>

image

<br>

 ### Tablet

image

<br>

image

### Mobile

<br>

image

<br>

image

## Sketch file (new)
[032421_CarbonPatterns_IconTable.sketch 4.zip](https://github.com/carbon-design-system/carbon-website/files/6267009/032421_CarbonPatterns_IconTable.sketch.4.zip)

@thyhmdo For future ref this is where these enhancements go and how they get spec'd out more or less.

### LINKED ISSUE #2366: Followup work for status indicator pattern
Follow up for #2294, closes #2199

- [ ] add the tables (# https://github.com/carbon-design-system/carbon-website/issues/2199)
- [ ] could you also add the new colors into the palette component (I need the blue, purple and gray added) — I'm hoping this won't be too much trouble, @josefina already has these components made... I just can't change the colors in markdown. https://www.carbondesignsystem.com/data-visualization/color-palettes
- [ ] I also need to enter the yellow and orange values into the palette component, but I do not need it to have tabs... it doesn't have separate dark and light values.
- [ ] Enable sidebar link to pattern

image

Light values

image

Dark values

image

image

image

vpicone:
closes #2366
closes #2199

TODO:
- [x] create table constituent components
- [x] add the rest of the yaml
- [x] color pallet stuff
- [x] glyphs
- [x] enable sidebar link
- [x] cleanup styles
- [x] address naming questions and fix filenames


vpicone:
@jeanservaas as I'm goin through adding the data:
- Should `Caution` (base) come before `Caution Inverted` (variant)?
- Should the token for `Caution` just be `$support-caution` (why have a support-caution-alt without a plain support-caution)?

Naming conventions:
- We should normalize the names for variants: `inverted` vs `alt` vs `square` etc.
- Symbol vs Shape vs Glyph vs Small? Maybe we should pick one and use it everywhere to indicate the compact variant.

aagonzales:
Here is what i'm recommending for the token names.


Token | Usage | White | Gray 10 | Gray 90 | Gray 100
-- | -- | -- | -- | -- | --
$support-01 (v10) </br> $support-error (v11) | Critical, warning, misuse | Red 60 | Red 60 | Red 50 | Red 50
$support-caution-minor | Status minor caution | Yellow 30 | Yellow 30 | Yellow 30 | Yellow 30
$support-caution-major | Status major warning | Orange 40 | Orange 40 | Orange 40 | Orange 40
$support-undefined | &nbsp; | Purple 60 | Purple 60 | Purple 50 | Purple 50
$support-02 (v10) </br> $support-success (v11) | Success | Green 60 | Green 60 | Green 50 | Green 50
$support-04 (v10) </br> $support-info (v11) | Informative content | Blue 70 | Blue 70 | Blue 50 | Blue 50

jeanservaas:
@vpicone I'm not sure what's going on with the colors here. I have the sketch file open and the values for the svg assets I made seem to be correct

### High attention color value errors

- The dark theme red values should be Red 50 (This should be the #support-01 token for both light and dark themes)

**Spec:**
image

**Preview:**

- In your preview it looks like the values are alternating between a Red 50 and a Red 40 (or 30?)

image

### Medium attention color value errors


**Undefined Spec:**
- The light theme purple value should be Purple 60 and the dark theme value should be Purple 50 (there is no token for this at the moment)

image

**Undefined Preview:**

- It looks like both light and dark values are the same and both look like purple 40

image


**Checkmark Spec:**

- The light theme green value should be Green 60 and the dark theme value should be Green 50 (This should be the #support-02 token for both light and dark themes)

image


**Preview:**

- It looks like both light and dark values are the same and both look like green 40
image

**All blue status icons spec:**

- The light theme blue value should be Blue 60 and the dark theme value should be Blue 50 (This should be the #support-04 token for both light and dark themes)

image

**All blue status icons preview:**

- This doesn't even look like our blue and again both values are the same

image

**Pending spec:**

- The light theme gray value should be Gray 70 and the dark theme value should be Gray 30 (This should be the #icon-02 token for both light and dark themes)

image


**Pending preview**

- Both light and dark values look the same and both look like gray 40 or something.

image


### Low attention color value errors

- See above fix blue and grays accordingly

jeanservaas:
@vpicone 

I'm keeping this separate from the visual updates, these are the responses to your questions from above:

> > 
> Should Caution (base) come before Caution Inverted (variant)?
> 

No because we’re going down from maximum severity


> Should the token for Caution just be $support-caution (why have a support-caution-alt without a plain support-caution)?

Anna has updated the token names below, the ones I had in Notion were for placement only as she worked on them. Remember, these are not the new v11 token names that Anna has been working on. This is tricky because all of these token names will be updated, (i.e. $support-03 will become $support-caution-minor will v11)

image

What I do need to figure out, @aagonzales, is how to treat the fact that the outlined caution icon is a different color. It may not get a token, we might just have to specify _$support-caution-minor, yellow 60_


> Glyphs seem kinda thrown in at the very end, can we fit them in better? Maybe as an even lower attention level (info?)?

Glyphs are not a lower attention level, but map to the status levels above, they are just smaller, in that sense they would not belong above with the icons. They are good where they are a status variant.

### Naming conventions:

> Should we name based on purpose instead of shape?

We are definitely trying to do that, however, we're also at the mercy of some inconsistent icon-naming conventions that have already been established on the Carbon site. The two most inconsistent names here I think are "Circle dash" and the "Warning hex" (because Circle dash doesn't speak to purpose and Warning hex is one of the only ones that specifies shape... like, there's no "Checkmark circle") 

Both are the names of the icons right now on the site, so the difficulty is making sure people can find them via their icon names on the site. Otherwise, you could call "Circle Dash" "Not started" and "Warning hex" "Warning alt" (since there is not distinct usage...); I suppose you could call the uncontained checkmark "Checkmark alt" too. But we'd have to try to rename these assets on the site.

> We should normalize the names for variants: inverted vs alt vs square etc.

See above, we are attempting to standardize this convention... but we're talking about more than just the status indicator page, we're talking about the entire icon library


> Symbol vs Shape vs Glyph vs Small? We should pick one and use it everywhere to indicate the compact variant.

This is unfortunately the result of losing Jan mid stream before I could get a comprehensive review off of this one. I attempted to standardize in that "Symbols" or "Glyphs" are the  icons used WITHIN the shape or container. Both large and small usage statuses have shapes... the difference is, the small usage ones ONLY have shapes. I definitely don't want to call them "symbols" or "glyphs" for this reason. They could be called "small usage indicators," instead I suppose.

> When should glyph include shape vs name (Critical vs Circle)

This is a good point but tricky here... part of the reason that I'm nervous about assigning a name like "success" instead of "circle" to the status shape indicator, is that I'm worried that people will sometimes divorce the color from the shape (i.e. they may use multiple colored circles as long as they have type next to them in a table. I suppose we can outlaw that type of usage, and just change "Circle" to "Stable" and "Square" to "Information" (square is harder).

> When should glyph indicate filled? Do we need to if there's no outline version? (Circle/square)

There are no outlined versions of glyphs, so we don't want to indicate that. Let's call "Circle stroke" "Draft"

> Undefined is usually a bad name for things in programming, it's a "reserved word" so you can't name variables with it (like function, if, else, etc.)

Again it's just the name of the status icon. It indicates what about "unknown" or "outlier?"

image


### Usage questions:

> When do you use alternates?

This is what I say about when to use alternates, many teams request them, this will also ease a transition to a standard:
(I could change the title to "Alternate status icons"

_**Alias status icons**
In several cases, we offer multiple shape options for one type of status indicator. For example, ‘warning’ ‘help’ and ‘information’ status icons have multiple variants to convey similar or exactly the same information. Often times, users of certain products have grown accustomed to say, a hexagon instead of a circle to convey a critical warning. Or a team perhaps wants to go the extra mile and offer yet another differentiator for accessibility. Although consistency is always the goal, there’s no need to avoid differentiation to accommodate user preferences. Whichever style you choose however, avoid mixing throughout the UI.
If an alias that your product frequently uses has been removed from the set and you can make a case for the importance of including it here, please let us know._



> What is the difference between not-started and disabled?

Not-started is a step within a progress indicator component or a task that has not yet begun, this symbol is in wide use now, we are just accommodating it
 


### Illustration questions:

> Is this Caution glyph outline intentional?

image


Yes this is intentional for contrast accessibility; We want to preserve the yellow’s inherent meaning, rather than making the whole thing baby poop brown

mjabbink:
@jeanservaas Please remove the full return from the Overview
https://carbondesignsy-git-fork-vpicone-status-indicator-carbon-32987e.vercel.app/patterns/status-indicator-pattern/#overview

That should be one paragraph.


Screen Shot 2021-06-14 at 6 38 04 PM

Screen Shot 2021-06-14 at 6 37 51 PM

mjabbink:
@thyhmdo @jeanservaas The green caret is too small in this sample. If this icon were to be used in this situ it would need to be larger.

Screen Shot 2021-06-14 at 6 51 26 PM

mjabbink:
@jeanservaas @thyhmdo Looks like some reds are too light in some of the dark theme icons. The top row looks correct and the ones underneath are lighter in the stroked versions and the last fill version  of “misuse”

vpicone:
@jeanservaas Sketch SVG exports are the worst. They have some weird filter being added? I can hand fix these, but if possible illustrator exports are generally much more clean.

vpicone:
@jeanservaas follow up on the SVG structure. A lot of them use #161616 for internal marks on dark theme. Should they instead just mask the icon so background is visible through the

vpicone:
@jeanservaas ready for visual review again. I'll take a look at the content stuff tomorrow

jeanservaas (commented):
I made some updates to the YAML and committed, just a couple of other things.

- Dark filled check is bigger than other icons, all should be 20px

image

- Somehow the type in all the tables is Gray 70, it should all be Gray 100

image

- I didn't try to change order in the YAML, but could you put the  Major caution orange icon above the Yellow Minor caution... and do the same with the shape indicators? I hear your concern, but they go down from severity, which is why the orange is first.

vpicone:
@jeanservaas the text color is coming from the `StructuredList` component up stream. Do we want it to be gray 100 there as well or just have a one-off here.

vpicone:
@jeanservaas ready for another review

mjabbink:
@jeanservaas @vpicone Looks good. Ready to go I think right @jeanservaas?

jeanservaas:
@vpicone 

Colors all look good now, only a few small things:


- Looks like Caution inverted phantom row is still there at the end, just need to remove it

image


- Could you make this image 8 cols instead of 12cols?

image

- I want to hulk smash two more type updates

jeanservaas:
@vpicone LAST THING! Yay!

I just hulk smashed a couple more things into the YAML... but I was afraid to change the order of anything in the YAML... thought I probably could've just done this too... but for the small guys, I had to change the name to 'Major caution' and 'Caution' too ... and could you swap their order? just like we've done with the icons above?

image

PR #2383, opened 2021-06-04
docs: v11 size token updates for pagination component

thyhmdo:
This PR updates the size tokens on the Style tab for the pagination component.

### DIFF
--- a/src/pages/components/pagination/style.mdx
+++ b/src/pages/components/pagination/style.mdx
@@ -32,12 +32,18 @@ it's being paired with.
 
 | Class                               | Property                    | px / rem  | Spacing token |
 | ----------------------------------- | --------------------------- | --------- | ------------- |
-| `.bx--pagination`                   | height                      | 48 / 3    | –             |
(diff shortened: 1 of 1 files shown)


emyarod (approved):
looks good to me

vpicone (commented):
Sorry, I haven't been in the size prop loop, but we only have one size right? Is there a reason we're calling it "Large" and not "Normal" What is "Large" relative to?

dakahn (approved):

aagonzales:
You're not out of the loop @vpicone we're just trying to figure these things out on the fly. I agree though, if a component only has one size then it doesn't need a size section. `height` should just be indicated in the structure table and image.

aagonzales:
So I think we can actually just close/delete this PR since no changes are actually needed?

jillianhowarth (approved):

aagonzales:
Closing this PR, no changes needed sense it doesn't have size variations.

PR #2410, opened 2021-06-17
docs: typography for expressive and productive

janchild:
This PR is for the new guidance for expressive and productive type sets. 

Issue for component work: https://github.com/carbon-design-system/carbon/issues/8934

#### Notes from Scott:

- I took the `typeScale` and `typeSets` JSON and moved them into their own individual files to make `TypesetStyle` more readable.
- Lots of renaming style tokens to match the spec. Pretty much everything got a renaming.
- Refactored `TypesetStyle` to allow for a boolean prop to turn the old header off and on. Even though there's no use-case for the old header, I didn't have the heart to remove the code, thinking we might bring it back someday. Added defaults of `breakpointControls={false}` & `header={false}` to make the component easier to implement.


jeanservaas (commented):
## Type sets tab comments

- The legal -02 should be the same value as the label 02

> Type: IBM Plex Sans
> Size: 14px / .875rem
> Line height: 18px / 1.125em
> Weight: 400 / Regular
> Letter spacing: .16px
> Type set: Expressive

image



- Can we force stack these fixed headings to show the line height change? Like this:

image


**Just for these:**

> $fixed-heading-compact-01
> $fixed-heading-compact-02
> $fixed-heading-01
> $fixed-heading-02





- This is also wrong on our site now, but both the Quote styles area actually supposed to be Plex Serif (not Sans)

image



- I’d remove these warnings from the table (for all four Display sizes) and add **one** warning to the paragraph above instead:

image


## Type usage tab


Can we move this last paragraph “For more details about…” above the image?

image


I understand why it’s there, but I think the type hierarchy, with it being awkwardly right under the caption of the image, is hurting the read, even though the placement makes sense.

janchild:
@sstrubberg Also, I have one edit to the `Small styles` table. 

For `label-01` and `label-02` could you please change the text to the following:

> This is a multipurpose type style that can be used for component field labels, helper text, error messages, as well 
> as captions. It should not be used for body copy.

janchild:
@sstrubberg, here is the final list of type tokens. 

We had a review of all the content today and everything looks good. We have changed `Small styles` to `Utility styles` so if you want to make any changes on your side to match the new name for the littlies, I guess now would be the time. 8-) 

[Typography type token list.xlsx](https://github.com/carbon-design-system/carbon-website/files/6956640/Typography.type.token.list.xlsx)

laurenmrice (changes requested):
@sstrubberg 

I was going to try to make comments inline to make it easier for review, but all of my comments are about the actual tables on the page and it won't let me do that.

---

### Type sets page

**legal-01** and **legal-02**
- In the example text for these tokens, I would reword them to match the other type token example text. We don't need to repeat the token name in each example.
- For `legal-01`, say this instead "This is for legal copy appearing in product pages."
- For `legal-02` say this instead "This is for legal copy appearing in web pages."

**body-01**
- I would change the first part of the first sentence of the example text to say "With a slightly taller line height than body-compact-01...". This will be more consistent with the example text in the table below for body-02.

- In the last sentence of the example text, Body-long-01 should say "Body-01".

**body-02**
- In the example text, body-short-02 should say "body-compact-02".

**Fixed headings sections in table**
- All text and inline code snippets showing fixed heading type token names should remove the `fixed-` part of the name.
- For example, fixed-heading-01 should say "heading-01", etc.

janchild:
> @sstrubberg
> 
> I was going to try to make comments inline to make it easier for review, but all of my comments are about the actual tables on the page and it won't let me do that.

Thank you, @laurenmrice! Great edit. 100% agree!

PR #2431, opened 2021-06-29
docs: add feducation video to v11 page

mattrosno:
Adds FEDucation video to the v11 page to make that replay more easily available, especially for Carbon users outside of IBM.


mattrosno:
@vpicone I don't see any docs for how to add playback tracking to the Vimeo component. (So somebody could rewind and watch portions again, fast forward, etc.) Is there a secret way to do this?

https://gatsby.carbondesignsystem.com/components/Video#props

image

joshblack (approved):

vpicone:
> @vpicone I don't see any docs for how to add playback tracking to the Vimeo component. (So somebody could rewind and watch portions again, fast forward, etc.) Is there a secret way to do this?
> 
> 
> 
> https://gatsby.carbondesignsystem.com/components/Video#props
> 
> 
> 
> image
> 
> 

I think those setting happen on the Vimeo side.

jillianhowarth (approved):

sstrubberg (approved):

Issue #2435, opened 2021-07-02
[pattern] status indicator grid issues


jeanservaas:
### Grid issues:
Also, the the status icon and status shape tables are not on grid, to the spec that was given. And they're not consistently off grid either. 

**See misalignment:**

![image](https://user-images.githubusercontent.com/43144260/123987918-95b5b480-d995-11eb-96cf-4b4e624686d6.png)

**Table specs:** (See complete table specs at breakpoints here)
https://github.com/carbon-design-system/carbon-website/issues/2199

To do:

- @vpicone we need to align the tables to the specs that were in the other issue; I didn't catch these discrepancies in my review

mjabbink:
Let’s not forget about this grid, alignment fix. This should be addressed in the next sprint.

mjabbink:
@alisonjoseph @joshblack Please make sure this is added to this sprint. This one can't slip through the cracks.

alisonjoseph:
@mjabbink @jeanservaas I'll get @vpicone to chime in here also, but from my understanding this is using the Structured List component from Carbon which is _not_ grid-aware. It would likely require a complete re-write of that component or a new one to be able to figure out a way to get this on-grid. That being said, I see the issue especially where the different tables don't align to each other, and wonder if there's something we could do to make it a little better until we have time to prioritize a new component?

mjabbink:
@alisonjoseph @vpicone If the first two columns of icons are the same shouldn't those at least be the same in each structuralist? maybe as a first fix, those icons columns could be the same which would make a difference.

another issue
@alisonjoseph @jeanservaas @vpicone — I discovered in mobile the dark theme column is not visible or turned off. Should that be a separate issue or keep it here?

<img width="392" alt="Screen Shot 2021-08-04 at 11 21 52 AM" src="https://user-images.githubusercontent.com/32881239/128217790-057ef037-63ff-405d-a03d-ad54ee7fdae7.png">

mjabbink:
@alisonjoseph @joshblack @aagonzales Please include this in the next sprint.

mjabbink:
@alisonjoseph @sstrubberg — Please resolve this issue and prioritize. 
@jeanservaas Please raise these types of issues before 4 months go by.

jeanservaas:
@alisonjoseph @sstrubberg @mjabbink 

[...]

PR #2436, opened 2021-07-02
Status indicator follow-up

### LINKED ISSUE #2430: [pattern] status indicator content fixes
### Content issues:

The Security team pointed out that their guidance around data decorator components directly conflicts with some of what was in the shape status indicator guidance. Their shape status indicators are widely deployed in product and this discrepancy will cause confusion. Changing their shapes to conform to our guidance would also be very costly. We've decided to fold in their risk/severity scale shape indicators into our shape indicator lexicon as they are.

**Security guidance links:**

 https://ibm.box.com/s/mb2dwp56k4rvgl50kwr9tniuqlpbpis7
https://ibm-security.carbondesignsystem.com/?path=/story/components-security-decorator--icons

# To do:

- @conradennis is helping us add two shapes to the library that we're currently missing (the 'Critical' right triangle' and the 'Low severity' square with a stroke for contrast accessibility)
https://github.com/carbon-design-system/carbon/pull/8960

- @vpicone we also need to update the warning major and warning minor assets for accessibility

image

- @vpicone we need to add three status shapes into the table I will give you the svgs and the hex

image

- We need to update the table content in the YAML file (I have content @vpicone if you open the PR and fix the tables, @jeanservaas  can update the YAML in the issue)

- @jeanservaas needs to update the attached images to reflect the real Security product UI (Including the legend image, it's an incorrect example of a legend).

Full new status shape table @cameroncalder for approval

Status icons_Update

- We also have to update the gray in the status palette 

image

----

**- Images to fix:**

Screen Shot 2021-06-29 at 12 25 27 PM

Screen Shot 2021-06-29 at 12 17 16 PM

Screen Shot 2021-06-29 at 12 17 34 PM

vpicone:
Closes #2430

related: https://github.com/carbon-design-system/carbon-website/issues/2435

- [x] palette fix
- [x] glyph assets added/removed/adjusted/modified


jeanservaas:
- Please also update the Caution major and Caution minor icons with the new assets: 

image


Should look like this with the stroke (darks need to be changed too because the exclamation mark is smaller):

image

- Please update the Gray status icons with the new assets I sent you over slack too (pending, help and unknown); the gray status value has changed, Mike wanted to go lighter with it. We just need to change them out

image


- Please make the gray value say "Gray 60" (not "gray 30) and make sure that the gray hex is # 6f6f6f

image

- Make gray value on the light tab say "Gray 50" and the hex value say # 8d8d8d (and change the color of the gray to # 8d8d8d — it's a gray 30 right now)

image

- I also sent you new Draft status shapes as we need to update the gray there too, Mike doesn't want to use the $icon-02 token anymore for the status grays, which is why we're updating it.

image

- Can you add this caption under the status shape table like in the spec

image

Here's the live text:

> The Security team has already established a severity/risk scale with associated shapes that is used widely in their products. To avoid disruption and confusion, we’ve incorporated the severity scale here as it exists in their product.

Just reattaching the status assets i added on Slack:


[New_status_assets.zip](https://github.com/carbon-design-system/carbon-website/files/6756538/New_status_assets.zip)

vpicone:
@jeanservaas I think you might have been looking at an old preview, could you check again and let me know what remains?

joshblack (approved):

jeanservaas (commented):
@vpicone thanks for making most of these updates. All of the pesky assets with strokes look good now. Only a couple of tweaks left. 

- Just need to update the outlined version of both the major and minor caution, they should be Orange 60 (BA4E00) and Yellow 60 (8E6A00) respectively but are lighter. I attached the assets here, forgot to add them before, sorry.



image

- One of the gray status icons (Unknown) didn't get updated to the new Gray 60 (6F6F6F) / Gray 50 (8d8d8d), we just need to update it to the right hexes to match the others

image

- Also can you add that caption under the status shape table about the Security indicators in italic

image

> The Security team has already established a severity/risk scale with associated shapes that is used widely in their products. To avoid disruption and confusion, we’ve incorporated the severity scale here as it exists in their product.



[major_minor_cautions.zip](https://github.com/carbon-design-system/carbon-website/files/6771783/major_minor_cautions.zip)

vpicone:
@jeanservaas what's the point of the caption? Is it temporary? If this is the color scheme/scale then this is the scale. Not sure why we'd need to add this caveat.

jeanservaas:
@vpicone I suppose you're right. I was not originally thinking of it as being temporary — it was more about basically having Security's severity scale kind of folded into our status lexicon. It felt weird to give it it's own section, so I was kind of captioning it in that way.

I guess I'm OK with just leaving it out, or calling out that it belongs to security within the table descriptions

jeanservaas (approved):

Issue #2440, opened 2021-07-07
Theming issue with CopyButton on CodeSnippet demo


tw15egan:
Rel https://github.com/carbon-design-system/carbon-website/pull/2433

Noticed an issue with the `CodeSnippet` copy button

![Screen Shot 2021-07-07 at 11 07 41 AM](https://user-images.githubusercontent.com/11928039/124784367-dc665a00-df13-11eb-940d-3b31b9d82d7b.png)

PR #2433, opened 2021-06-30
fix(Typography): fix issues with inline copy button

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/2432

Fixes some issues on the Guidelines --> Typography page

#### Changelog

**New**

- Ensures the inline code snippet copy text is properly hidden by ensuring the `copy-button` styles are present. 
- Added the `light` prop to the code snippet

**Changed**

- Synced `carbon-components`  and `carbon-components-react` packages, and updated a few other out of date packages

**Removed**

- Removed the `-` between `Line height` and `Letter spacing`

**Testing**

Go to Guidelines --> Typography --> Productive / Expressive and ensure the inline code snippet has a visible background, there is not double `Copied!` text, and that there is no hyphen between `Line height` and `Letter spacing`

PR #2441, opened 2021-07-07
fix(CodeSnippet): use token for copy button

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/2440

Ensures `CodeSnippet` copy button uses the correct token

#### Changelog

**New**

- Sets the background color to the correct token after importing the `copy-button` styles to fix #2433 

**Testing**
- Go to the Codesnippet code demo and ensure the copy button changes with the themes. Also, go to Guidelines --> Typography --> Productive and ensure the copy button is styled correctly and does not show duplicate `Copied!` text


jnm2377:
Should the copy btn look like this on hover? Seems like maybe the icon color should change too? but also not sure if it's the right hover color
Screen Shot 2021-07-07 at 12 01 48 PM
Screen Shot 2021-07-07 at 12 01 39 PM

tw15egan:
@jnm2377 oooo right, good catch! I'll need to dig in and see what else needs to be set 👍🏻

jnm2377 (approved):

joshblack (approved):

PR #2444, opened 2021-07-09
Color implementation tab

aagonzales:
## Ready for review:

New tab for color implementation to explain:
- Layering model tokens
- Inline theming 
- Light or dark mode

**Preview link:** 
https://carbondesignsystem-git-fork-aagonza-f5b8b0-carbon-design-system.vercel.app/guidelines/color/implementation

**Questions:**
- Do we need to list the tokens out on this page or is it ok to link out to them? Where should contextual tokens link out too?
- What order should the tabs go in?
- Are the full UI examples in layer tokens section too big?


joshblack:
@aagonzales when looking at the implementation page, one thing I was curious about was if the two layering models could be presented one after the other instead of at the same time.

The explicit layer tokens seem like they'd be familiar / commonly understood and would work for most use-cases. Then, we could introduce contextual tokens as a way to make work reusable across layers.

I was curious what you thought about this / if you had explored it already and what your thoughts were 👀

aagonzales:
@joshblack Do you mean not grouped in the same bucket of "layering model"? They are in their own buckets inside that, they each have their own H3 section. Are you suggesting they each just be H2 and remove the info around layering model and just introduce them as independent topics? 

I have them grouped together because I see it as an either/or situation to apply the layering model. I haven't really explored making them independent sections that much because in my head they are related. I could be convinced otherwise if you want to explain/show what you're envisioning.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+
+</AnchorLinks>
+
+## Layering model tokens
+
+There are two ways to implement the
+[layering model](/guidelines/color/overview#layering-model) within a single
+theme, either by using explicit layer tokens or smart contextual layer tokens.
+Both ways are acceptable with different pros and cons that should be weighed in
+deciding which to use when building your product.
```
joshblack, line 30: One attempt at this intro paragraph:

> Explicit and contextual layer tokens are used to implement the layering model in a theme. These techniques can be used independently in a product or can be used together.
>
> Explicit layer tokens are standard design tokens. Each layer token corresponds to a specific layer on the page and can be used exactly like any other token from Carbon. Contextual layer tokens are a special case of tokens where the value changes depending on where the token is used on a page. This type of token is incredibly useful for building reusable components that work across layers.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+
+| Token type                                                                          | Definition                                                                                                                                                                                             |
+| ----------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| [Layer set tokens](/guidelines/color/implementation#layer-set-tokens)               | Explicit tokens used to manually map the layering model onto components. They come in sets based on their level in the layering model and are part of the theme package.                               |
+| [Contextual layer tokens](/guidelines/color/implementation#contextual-layer-tokens) | Abstracted tokens that automatically infer value based on the context a component inside the layering model. They are part of the component package and don't have an explicit value assigned to them. |
```
joshblack, line 35: > They are part of the component package and don't have an explicit value assigned to them.

I would just drop this line when talking about which package they belong in. There are several ways people could get access to them so may not be helpful specifying a single package.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+
+<Row>
+<Column colLg={12}>
+
+![Layering model token comparison](images/color-implementation-compare.png)
```
joshblack, line 40: Love this image 💯

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+| `$border-subtle`          | `$border-subtle-01`          | `$border-subtle-02`          | `$border-subtle-03`          |
+| `$border-subtle-selected` | `$border-subtle-selected-01` | `$border-subtle-selected-02` | `$border-subtle-selected-03` |
+| `$border-strong`          | `$border-strong-01`          | `$border-strong-02`          | `$border-strong-03`          |
+--->
+
+<br />
+
+### Layer set tokens
+
+Layer tokens are sets of explicit tokens used to manually map the layering model
```
joshblack, line 73: Since these are typical design tokens, I think we could drop calling them explicit tokens and refer to them as design tokens. This could help make the concept more familiar.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+onto components. They are part of the theme package and come in predefined sets
```
joshblack, line 74: I would mention what package they sit in as this can be confusing. For example, if someone is using the `@carbon/styles` or `@carbon/react` package this would be available to them. They may read this and assume they need o install the theme package too.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+There are four sets of color layering within a theme: base layer, layer 01,
+layer 02, and layer 03. After each layer of color added you move up to the next
+layering level and use the next set of layering tokens. For example, on top of
+`$layer-01` you would place `$layer-02`. These layer set tokens are identified
+by a number suffix (-01, -02, -03) attached to the core token name, such as
+`$field-03`. The number indicates which set the token is a part of. _Note:
+previously in v10 many color tokens had numeral suffixes, now in v11 only
+layering tokens will have this distinction._
+
+Not all layer set tokens use `$layer` as its token name. Token sets also include
```
joshblack, line 87: Seems like we're saying "layer set tokens" and then "token sets". Are they two separate things or should they be the same term?

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+- Layer set 03: Cyan
+
+<Row>
+<Column colLg={8}>
+
+<Tabs>
+
+<Tab label="White">
+
+![Layer set tokens applied to dropdown in the White theme](images/color-implementation-layer-set-W.png)
```
joshblack, line 113: Great example 🔥

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+</Tabs>
+
+</Column>
+</Row>
+
+<br />
+
+#### Building components with layer tokens
+
+Building component with layer tokens works much like how you would have built
```
joshblack, line 206: Could the v10 references be better located under migration documentation from v10 to v11? Then this page could make the assumption that people are doing things in v11 only.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+</Column>
+</Row>
+
+<Caption>
+  Example component spec in the white theme using the layer set 01 tokens.
+</Caption>
+
+### Contextual layer tokens
+
+Contextual tokens are smart tokens used to automatically map the layering model
```
joshblack, line 232: Calling this new thing both contextual tokens and smart tokens can be confusing given that we don't really say what they are. Maybe we could describe contextual tokens as:

> Contextual tokens are a special case of design token where the value changes depending on where it is used on a page.

And then refer to them as only contextual tokens throughout the page?

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+### Using the layer component
+
+The layer component is used with the contextual tokens to implement the layering
+model in code. The layer component and contextual tokens are essentially a short
+cut for development because the layer component will apply the correct values to
+a single component without having to develop explicit hard coded variants.
+
+<InlineNotification>
+
+**Coming soon:** Developer documentation is currently be worked on. Check back
```
joshblack, line 372: Some Layer dev docs that we have up: https://carbon-react-next.netlify.app/?path=/docs/components-layer--default

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+
+</Column>
+</Row>
+
+### Using the layer component
+
+The layer component is used with the contextual tokens to implement the layering
+model in code. The layer component and contextual tokens are essentially a short
+cut for development because the layer component will apply the correct values to
+a single component without having to develop explicit hard coded variants.
```
joshblack, line 368: Use the `Layer` component in order to have contextual tokens map to the correct layer value. Each token that is used to build a component rendered in a `Layer` component will automatically map to the next level based on the current layer set. These components can be nested up to three times, the last level corresponding to the 04 layer set of tokens. By default, tokens use the 01 layer set of tokens.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+![Example of inline themes](images/color-implementation-inline-2.png)
+
+</Column>
+</Row>
+
+### Using the theme component
+
+<InlineNotification>
+
+**Coming soon:** Developer documentation is currently be worked on. Check back
```
joshblack, line 429: Theme dev docs for v11: https://carbon-react-next.netlify.app/?path=/docs/components-theme--default

joshblack (approved):
Looking great! The images are so helpful 🔥  Left a couple of comments/suggestions, not sure how helpful they are so please ignore them if they are unclear or unhelpful haha.

src/pages/guidelines/color/implementation.mdx
```diff
@@ -0,0 +1,546 @@
...
+  In IBM Cloud (beta), the theme control can be accessed through the user
+  profile panel in the header (left). The "Change theme" link triggers a model
+  where the user can save their theme preference (right).
+</Caption>
+
+#### Inline themes
+
+Mixing themes inline is still allowed with light or dark mode. Mixing inline
+theme contrast between elements in different modes is also allowed. It is very
+common for products to have side panels or UI Shell elements be high contrast in
```
jillianhowarth, line 515:
```suggestion
common for products to have side panels or UI shell elements be high contrast in
```

jillianhowarth (approved):
This is really awesome, Anna! As someone who has been very out of the loop on the token work, I thought it was super thorough and totally made sense and was easy to follow. The suggested changes are mostly just small fixes. Great work!! 🎉

src/pages/guidelines/color/implementation.mdx
aagonzales: We don't really use the term "design tokens" anywhere else in our system. At least not in v11. We can introduce it though we just need to explain what it is somewhere and if/how it's different than just a plain "token."

src/pages/guidelines/color/implementation.mdx
aagonzales: I thought that was the big difference between the two (if im remembering previous conversations correctly)? So knowing which package they're in is irrelevant?

src/pages/guidelines/color/implementation.mdx
joshblack: Token as a term totally makes sense to me too 👍  I thought they were interchangeable but totally get if they're not. If Token is used/has been used then using that instead of design token makes sense to me.

src/pages/guidelines/color/implementation.mdx
aagonzales: "smart tokens" was more a metaphor then an actual term I was using to help people understand how they're different than explicit tokens. Jan used this term a lot when talking about them so I thought it was a safe term.

src/pages/guidelines/color/implementation.mdx
aagonzales: @joshblack Do we ever want to have those docs in the storybook on the website? Or should we just link out to them and not has a specific section for dev docs on this page?

src/pages/guidelines/color/implementation.mdx
aagonzales: Again to we want to embed any of these docs on this page or just always link out to storybook?

src/pages/guidelines/color/implementation.mdx
aagonzales: Ok I see them as interchangeable too. When ever Jan talked about these tokens he referred to them as explicit vs smart so that's where these terms came from. It made sense too me but maybe seem like official name instead of just adjectives.

src/pages/guidelines/color/implementation.mdx
joshblack: Right, since packages can re-export items from another package it's not necessary to know which package they may originally be defined in. Instead, a developer needs to know how to get it from the package that they're using.

For most teams, the hope is to use `@carbon/react` directly and consume carbon through this one package.

src/pages/guidelines/color/implementation.mdx
joshblack: I'm down for either, whichever makes more sense! Just wanted to share what we had in storybook.

src/pages/guidelines/color/implementation.mdx
joshblack: I could definitely see how they're helpful. It seems like in my reading I thought of contextual tokens as the same as smart tokens which made the sentence feel redundant. However, it seems like you're meaning for it to be read is that contextual tokens are a type of smart token and that smart tokens are a broader group than contextual tokens.

src/pages/guidelines/color/implementation.mdx
aagonzales: OK i think we'll just link out to storybook for now since its easy and if we want to incorporate dev docs more intermittently like we do on the grid page we can update it later.

src/pages/guidelines/color/implementation.mdx
aagonzales: Yeah in the next round of edits I'll try to move some of the v10 vs v11 content out of here and into migration. That's a good idea.

Issue #2448, opened 2021-07-16
Searching for an icon and scrolling down causes the page content to disappear 


ud-ux-thomas:
## Summary
On the icons page, when I search for some icons, then scroll down, the page goes blank.

## Relevant information
1. On an intel-based Mac with all recent updates applied...
2. Using any of the following browsers: Firefox developer edition (latest), Firefox (latest), Google Chrome (latest) or Safari (latest)...
3. Go to the [icons page](https://www.carbondesignsystem.com/guidelines/icons/library/)
4. Search for "circle" (do not include quotation marks).
5. Scroll down using the touch pad, mouse wheel or arrow keys on keyboard.
6. Note that all page content is no longer visible.

I have attached a screen recording:
https://user-images.githubusercontent.com/38993885/125917614-3c87b29f-4d11-47cc-bd33-5aea61f3561e.mov

johnbister:
@ud-ux-thomas Confirming that I get the same bug in Chrome, Firefox, and Safari.

cc @carbon-design-system/developers Anyone know what's causing this issue?

tw15egan:
Also seeing this behavior, not sure what's going on either. @vpicone any ideas?

ballefar:
could it be related to dynamic imports and unstable network on client?

theoshu:
It's an error with the critical-severity/32 icon.

![image](https://user-images.githubusercontent.com/4643524/127492520-c4530f26-ebd1-4490-ad1a-996db5d474f7.png)

tw15egan:
The root of the issue was that some icons are only available in `glpyh` format, and those do not have a `/32` file causing the error. Added a check to see if an icon is glyph-only first before trying to import.

PR #2457, opened 2021-07-22
fix(IconLibrary): fix issue with glyph only icons

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/2448

Fixes an issue where we were importing all icons at the `32px` size, even if it was a glyph-only variant. This was causing the `guidelines/iconography` page to crash when one of these icons were loaded

#### Changelog

**Changed**

- If there is only 1 icon size, and the icon size is `glyph`, sets a flag that will load from the `index` file rather than `32` 

**Testing**

Go to `guidelines/iconography` and search for `caution` and ensure the page does not crash

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "dev": "gatsby develop -H 0.0.0.0",
     "dev:clean": "gatsby clean && yarn dev",
     "clean": "gatsby clean",
-    "build": "node --max-old-space-size=4096 ./node_modules/.bin/gatsby build",
+    "build": "node --max-old-space-size=8192 ./node_modules/.bin/gatsby build",
     "build:clean": "gatsby clean && yarn build",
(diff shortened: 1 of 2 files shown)


emyarod (commented):
it looks like there are some build errors with the new imports

theiliad (approved):

tw15egan:
@joshblack not sure whats going on with this, it all runs fine locally but seems to run out of memory when Vercel tries to deploy

joshblack:
@tw15egan ugh, I see what you mean. If you change this line: https://github.com/carbon-design-system/carbon-website/blob/main/package.json#L17 to use 8192 instead of 4096 does that help?

jnm2377 (approved):
Page doesn't break for me when I search for caution ✅

laurenmrice (commented):
Looks good to me.

PR #2504, opened 2021-08-23
Term 'light' used instead of 'dark'

esmahh:
In the paragraph 
'Dark themes
Gray 90 components are paired with Gray 100 backgrounds to make the Gray 100 Theme and Gray 80 components with Gray 90 backgrounds to make the Gray 90 Theme. Some components, such as buttons and toggles, share the same colors across both light themes.' - term 'light' (word behind the last one) is used and there should be 'dark'.

Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


laurenmrice (approved):
✅

PR #2531, opened 2021-08-31
Update to include required labeling guidance

jjennings7:
Added guidance for marking form fields required or optional depending on use case, and added Default values sub-section in the Offering Help section.

Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


kingtraceyj:
a few image updates:

**1. Form examples**
- change the background to grey20 instead of using both white and grey10
- try and stick to [aspect ratios](https://carbondesignsystem.com/guidelines/2x-grid/implementation/#aspect-ratios) for images
- could try a full 12 column image with the two forms side by side like [email templates](https://www.ibm.com/brand/email/templates#background-color) does

image

**2. Do/Don't image**
- can bump up to full width here so it's easier to see the required vs options denotions(12 columns)
- use an aspect ration for these

image

@jeanservaas any other suggestions here?

jjennings7:
@kingtraceyj Thanks for the image suggestions! I'm not actually sure how to go back in and replace the images – sorry, this is my first time contributing! Here are the revised images. Can you please replace them for me, or else direct me how to get back to the right place to do it myself? Thanks for your help!
For the form examples, I changed the background to grey 20 and made the ratios 1:1:
optional_example
required_example

For the DO and DON'T images, I added the grey20 background, and I tried 16:9 (12 cols) for them, but they're really tall so I'm not sure it's good to devote that much space to them:
DO_required_16x9
DON'T _optional_16x9

So, I also did a 1:2 version in case you'd rather keep them side by side (6 cols each to equal 12 cols/1:1 ratio total):
DO_required_1x2
DON'T _optional_1x2

dakahn (approved):
seems good to me 🏄🏾 the only thing that caught my eye was specifically calling out IBM products in the simple vs complex distinction. Is this not a general common distinction that we're just codifying/pointing out?

jjennings7:
@dakahn you're correct, it's a common distinction. I don't mind if you want to edit that out.

jjennings7:
@kingtraceyj please let me know if the image updates I provided above are acceptable. I'm sorry, I don't know how to go back into the markdown page to replace them. Thanks!

kingtraceyj:
@jjennings7 sorry for the delay on this — i will take a look today.

kingtraceyj:
@jjennings7 i'm thinking something like this for the do/don't image (plus the caption). i think we need a simpler form example so it's easier to see what's right and wrong about the image: 
image

kingtraceyj:
@jjennings7 i'm going to go ahead and approve the PR since the content is good to go. i'll open another PR and update the images with your approval.

kingtraceyj (approved):
approved. will open another PR or image up dates for the do/don't and use the same images from the component page for the form images.

PR #2552, opened 2021-09-23
fix: update link to the new URL of Gatsby theme

rodet:
As @vpicone recently wrote, https://gatsby.carbondesignsystem.com/ should be the new URL for the Gatsby Carbon theme. The old URL is still being used here, so I propose to use the new one instead.

#### Changelog

**Changed**

- The Carbon Gatsby theme URL link


kingtraceyj (approved):
looks good. thanks for the update!

sstrubberg (approved):

PR #2613, opened 2021-11-03
docs(disclosures): new pattern usage content

laurenmrice:
**New**

- Content
- Images


tay1orjones:
The files changed here will be somewhat inflated with formatting changes. Currently on the `v11` branch there are 16 files with invalid formatting against prettier. 

I pulled this down and ran `yarn format` which resulted in the formatting changes you see in this PR

jnm2377 (approved):
Wow. Great job on this! 🥇

src/pages/patterns/disclosures-pattern/index.mdx
```diff
@@ -0,0 +1,779 @@
...
+to take, whether it be upgrading an account plan or changing a location. Only
+add icons and links to a profile menu if completely necessary in order to not
+overwhelm the user and to reduce visual clutter within the menu.
+
+## Context menu
+
+A context menu is considered to be a disclosure pattern that is triggered by a
+right click interaction and opens a popover. A context menu is a defined list of
+options in a menu that appears above all other content on a page and is related
+to the selected content or element it was triggered from.
```
joshblack, line 426: I think the lack of a trigger button might make a context menu fall out of the disclosure distinction.

src/pages/patterns/disclosures-pattern/index.mdx
```diff
@@ -0,0 +1,779 @@
...
+<AnchorLink>Feedback</AnchorLink>
+
+</AnchorLinks>
+
+## Overview
+
+Disclosures support a wide range of different use cases in product interfaces
+and are commonly used to reveal more information or details about an element or
+content on a page. Disclosures can contain interactive elements while preserving
+accessible functionality.
```
joshblack, line 34: Maybe: "Unlike tooltips, the content expanded by a Disclosure may contain interactive elements."

I think the important thing to clarify is that if the Disclosure is two parts: trigger + content, then the content is what may have interactive elements without removing semantics like a tooltip would.

src/pages/patterns/disclosures-pattern/index.mdx
```diff
@@ -0,0 +1,779 @@
...
+
+At its core, a disclosure is comprised of two parts—a trigger that the user
+interacts with by clicking or using their keyboard and the popover that opens
+and discloses the content.
```
joshblack, line 38: At a high-level, I think this could be too narrow of a definition for a disclosure. [From the W3C](https://w3c.github.io/aria-practices/#disclosure), we have:

> A disclosure is a widget that enables content to be either collapsed (hidden) or expanded (visible). It has two elements: a disclosure button and a section of content whose visibility is controlled by the button.

And so this shows up in things like accordions, for example, that may not make use of a popover.

We can totally pair up a trigger button and a popover and use the disclosure pattern to toggle its visibility, but didn't want to say that only a popover may be used since the pattern is broader.

src/pages/patterns/disclosures-pattern/index.mdx
```diff
@@ -0,0 +1,779 @@
...
+
+To open a disclosure to reveal its content, click the trigger button it is
+related to. To close a disclosure, either click on the trigger button again,
+click anywhere outside of the open disclosure container or click on the close
+`x` icon if it is present within the disclosure.
+
+<Row>
+<Column colLg={8}>
+
+![Example of a disclosure opening and closing.](images/disclosure-opening-closing.gif)
```
joshblack, line 138: This example is one I'm not sure about on the disclosure side. I'm curious how we would expect focus to function in this situation and if it's more of a nonmodal dialog type thing

src/pages/patterns/disclosures-pattern/index.mdx
laurenmrice: So this is definitely true? If so, I am going to take this section for context menu out.

src/pages/patterns/disclosures-pattern/index.mdx
laurenmrice: Okay, I can take this visual example out or redo it. Does the written content for "Opening and closing a disclosure" make sense though?

jeanservaas (commented):
This looks amazing.

Couple of really small things:

These look a little horsey to me because I think it's scaled up like 1.5-2x. Could it be actual size or would that be really small? Actually, I just went through the rest of your images and it looks like they're all scaled the same way, so you probably shouldn't mess with the scale of this one.

It might seem a little less horsey if the content was styled like inline inputs where the label is the smaller 12px/14px type and the content after the label is the 14px/16px? Anyway, just thinking aloud on these. They're not that bad as is, just something I noticed might be a little more refined with the scale change (but it might also suck) 

image

Also, for callouts on gray 100 panel, you might want to use the Dark theme callouts, I don't think that magenta 60 is accessible on gray 100

image

jeanservaas (approved):

joshblack (approved):

PR #2614, opened 2021-11-04
fix: remove shadowed MediumPost component

alisonjoseph:
No need to shadow this, it should be pulled directly from the theme with the latest style updates. 

### Testing
- Ensure medium posts are still pulling in correctly on homepage and articles pages


sstrubberg (changes requested):
image

these are showing up in dark mode

alisonjoseph:
Good catch. 😬 I'll open a new PR to fix

PR #2635, opened 2021-11-11
v11[bugs]: assorted color token bugs on style tabs

aagonzales:
- updated `$field-disabled` instances to be just `$field`
- updated content switcher and inline loading v10 tokens
- updated/added border-bottom to disabled inputs as transparent


dakahn (approved):
🏄🏾

tay1orjones (approved):

PR #2642, opened 2021-11-16
chore(ci): add ci workflow action

tay1orjones:
In #2613 we found that some formatting errors had slipped into the v11 branch. Formatting and linting is only being checked via `lint-staged` in the git hook. There's a few ways this can be bypassed and invalid formatting can get into the build.

This PR adds a GitHub Action/workflow to explicitly check formatting and linting for every PR moving forward on `main` and `v11`, though it will only apply to ` v11` for now since we need to add this file to the `main` branch as well at a later time.

I can mark these status checks as required once merged in if preferred

#### Changelog

**New**

- add ci workflow for formatting check and linting

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: ci
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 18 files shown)


tay1orjones:
I also included the output of running `yarn format` against the current code in `v11`, so there's some formatting updates included in this PR alongside the workflow.

jnm2377 (approved):

aledavila (approved):

PR #2659, opened 2021-12-06
kit(figma): updated guidance

aagonzales:
Updated the getting started content and added images for reference to the Figma kit.


laurenmrice (approved):
🚀

PR #2718, opened 2022-01-27
fix: links on 2x grid overview page

laurenmrice:
**Removed**
- Removing links in the body paragraphs for Fixed and Fluid grid since they do not link anywhere and there is no respective section to link them to.


dakahn (approved):

thyhmdo:
Question, should those links stay and link to the bottom parts? 
image
image

if it's not the right way, let's pass it 🙌

laurenmrice:
@thyhmdo Further down the page those two sections already get an anchor link. Jeannie said we can just remove these first ones from this top section since it is not entirely clear.

thyhmdo:
sounds good

thyhmdo (approved):

PR #2737, opened 2022-02-15
Accessibility changes

mbgower:
Added 5 images and all new content to the Accessibility tab for code snippet.
Removed the Screen Reader testing material from the Usage page, and made one minor editorial that seemed acceptable during the playback.

Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


mjabbink:
@mbgower Thanks for tackling this! This is great. The next steps are for the design team to provide some feedback or do the design detailing on them illustrations and they need to span 8-columns in the page layout. Can you connect with @aagonzales to see who can support you on these improvements. Much thanks.

mbgower:
thanks @mjabbink 
Yeah, I wasn't sure if it was supposed to be 8. I'll adjust.
FYI, up until now we've been having these rewrites reviewed by CAG during handoffs.
@laurenmrice has looked over the ones we've done to date.

I thought I'd try this way to see if we could accelerate the changes. Here's what I'm doing this time:
- [x] propose to get access to do our own PR
- [ ] create a branch with our suggested changes
- [ ] walk through the PR as part of hand off
- [ ] if possible, implement the accessibility page as part of PR?
- [ ] potentially opportunity for knowledge share on rudiments of github/PR to a11y designers

src/pages/components/code-snippet/accessibility.mdx
```diff
@@ -19,79 +19,83 @@ import {
...
+</Column>
 </Row>
+
+The results of activating buttons are provided in text.
+
+## Developer considerations
+
+Keep this in mind if you’re modifying Carbon or creating a custom component:
+– the inline code text is implemented as a button so its text can be copied
+– single line snippets take an additional tabstop to support arrow key scrolling
```
tay1orjones, line 101: I think these need an empty line above them and a simple dash to get the proper list styling. Right now it looks like this 
image


```suggestion
Keep this in mind if you’re modifying Carbon or creating a custom component:

- The inline code text is implemented as a button so its text can be copied
- Single line snippets take an additional tabstop to support arrow key scrolling
```

tay1orjones (changes requested):

joshblack (approved):

tay1orjones (approved):

tay1orjones:
recheck

laurenmrice (commented):
@mbgower 
- I see the revised images! ✅
- Question: For any time we reference a keyboard function like `Space`, `Enter`, `Tab`, etc should we style them as code in markdown like we do on the Usage tab for the [Interactions section](https://carbondesignsystem-5mdpu7tkp-carbon-design-system.vercel.app/components/code-snippet/usage/#interactions) for consistency? We typically do this for all components.
- Also I understand that the word "Labelling" is technically correct and is the preferred spelling in the UK, while "Labeling" is the preferred spelling in the US. Not sure if it matters from a content perspective but wanted to note it here.

mbgower:
I will update the keys to make them like code.
It's funny you mentioned "labelling". I noticed that yesterday.
the specification uses aria-labelledby, etc.
I'll double check IBM Style and update non-technical uses accordingly.

laurenmrice (approved):
Looks good to me 🎉

mjabbink:
Seeing this now brings up wether mixing the two type of beginning and endings together. I did not think of that yesterday. I have a solution below that would allow (at least in this case) to maintain that little vertical straight segment and the beginning/ending of the large arch.

It requires some shifting to the right on the code snip and reducing the largest arch down.

Screen Shot 2022-03-02 at 5 25 23 PM

laurenmrice:
@mjabbink Thanks Mike, I will change this in the next iteration of Code snippet. 👍

mjabbink:
or you can divide the code field into 4 units and use that to place the arch?
Screen Shot 2022-03-02 at 5 41 37 PM

mjabbink:
or just use a small one and start the next tab from the center of the code field?

Screen Shot 2022-03-02 at 5 35 13 PM

mjabbink:
Not sure if the little straight segment is going to be a major pain all the time or only on occasion. Otherwise maybe the purple half circle might be better (as long as arrows are vertical or horizontal).

mbgower:
@mjabbink All great things to chew on. With a long horizontal target, there is more flexibility here than on some other components. As Lauren says, if we keep that in mind for other components we're updating images for, we can go back and choose which one seems to work better for other components

mjabbink:
@mbgower I recommend adding the “why” or “benefit”. I feel like if we say statement like this then we should follow up with why we do it or the benefit to the users.

Current
**What Carbon provides**
Carbon bakes keyboard operation into its components, as well as many other accessibility considerations. Could be an opportunity to educate.


Suggestion
**What Carbon provides**
Carbon bakes keyboard operation into its components, as well as many other accessibility considerations. This allows for... benefits, advantages... the why.

PR #2757, opened 2022-02-28
Fix v11 audit bugs: Progress indicator and Structured list

thyhmdo:
Related:

--------
**Progress indicator**
Style tab
- changed to `link-primary-hover` instead of link-primary

**Structured list**
Style tab
- changed to `layer-selected` instead of background-selected
- changed to `layer-hover` instead of background-hover


aagonzales (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #2772, opened 2022-03-09
Accordion accessibility

mbgower:
Population of Accessibility tab information


src/pages/components/accordion/accessibility.mdx
```diff
@@ -7,423 +7,88 @@ description:
...
-  </Column>
+<Column colLg={8}>
+
+![H1 and H2 annotations for headings, plus direction to ‘include accordion titles as headings'](images/accordion-accessibility-3.png)
+
+</Column>
 </Row>
 
-### Android screen reader tests
+If accordion titles act as headings, annotate for development.
```
aagonzales, line 70: Is this an image caption? If so it needs to be wrapped in the caption component.
```suggestion
<Caption> If accordion titles act as headings, annotate for development. </Caption>
```

src/pages/components/accordion/accessibility.mdx
```diff
@@ -7,423 +7,88 @@ description:
...
-    </StructuredListWrapper>
-  </Column>
+<Column colLg={8}>
+
+![annotation stating ‘position chevrons on the left of accordion titles'](images/accordion-accessibility-4.png)
+
+</Column>
 </Row>
+
+Annotate if the accordion chevrons should be left-aligned.
```
aagonzales, line 84: Another image caption? 
```suggestion
<Caption> Annotate if the accordion chevrons should be left-aligned. </Caption> 
```

src/pages/components/accordion/accessibility.mdx
```diff
@@ -7,423 +7,88 @@ description:
...
+
+## Developer considerations
+
+Keep these considerations in mind if you’re modifying Carbon or creating a custom component:
```
aagonzales, line 88: We try not to use contractions when possible. 
```suggestion
Keep these considerations in mind if you are modifying Carbon or creating a custom component:
```

src/pages/components/accordion/accessibility.mdx
```diff
@@ -7,423 +7,88 @@ description:
...
-  StructuredListBody,
-  StructuredListRow,
-  StructuredListInput,
-  StructuredListCell,
-  OrderedList,
-  ListItem,
-} from 'carbon-components-react';
+<PageDescription>
+
+Design annotations are needed for specific instances shown below, but for the standard accordion component, Carbon already incorporates accessibility.
```
aagonzales, line 12: Rephrase: It seemed off to me starting the page intro with a qualifier. 

> The standard Carbon accordion component already incorporates accessibility features, but further design annotations may be needed for specific instances shown below.

aagonzales (changes requested):
**Image feedback**: In general, these images don't seem to be using the new style updates we decided upon in our meeting last week.

The highlight around the first accordion item in this image appears to be off centered. 
image

"include" in the image should have a capital `I`, I believe with our style even statements start with a capital letter.
image


This image the scale of the laptop graphic to the page size doesn't seem correct. The page seems very zoomed in. Does this image need the context of laptop?
image

This image is very off-centered inside the image component.
image

mbgower:
@aagonzales 
> In general, these images don't seem to be using the new style updates we decided upon in our meeting last week.

Correct, I moved over the existing text, and @laurenmrice will be adding new versions of the images as part of her review.
This is mainly to inspect the text for issues, etc.

src/pages/components/accordion/accessibility.mdx
mbgower: The html `<caption>` element defines a table caption. It was our intent to use `<figcaption>`, but it didn't seem to be implemented in mdx files, so I left out for now.
I have an issue open on that

src/pages/components/accordion/accessibility.mdx
mbgower: @aagonzales The concern with that is if someone doesn't read the "but". It's more important that designers understand they _need_ to do annotations than that the standard one is hunky dory.

mbgower:
BTW @aagonzales I just noticed that you must have been autoassigned to this one. Lauren was all last week, so I assumed she would be this week too. I assume there's some kind of rotation. Why don't you hold off on reviews until I ping you that the new images are uploaded?

src/pages/components/accordion/accessibility.mdx
aagonzales: I disagree, it makes it feel like the user needs to do a lot more work but we actually provide most of the work out of the box. Most of the "design annotations" so far also seem like optional updates so they don't _need_ to address them. I think the current phrasing gives too much emphasis on the wrong intention, things you have to do yourself vs things Carbon provides. Just my opinion.

src/pages/components/accordion/accessibility.mdx
mbgower: BTW, we _really did_ have this called out already :)
image

aagonzales: hmm ok well we use that caption across all our pages on the website already. There spacing is off just enough that they don't really look like captions always when just using aregular text line. That's unfortunate.

aagonzales: Does it not work as a proper image caption if you put it inside the image component like:

```
<Row>
<Column colLg={8}>

![UI shell right panel example.](images/right-panel-style-1.png)

<Caption>UI shell switcher example</Caption>

</Column>
</Row>
```

aagonzales: @joshblack seems like a big problem is all this is true for the website. That means all our captions across the site are wrong.

mbgower: @aagonzales what I'm seeing is that, when captions appear under images, Carbon has tended to give the image the caption as both the alt and the title, then has just placed the caption underneath in a paragraph (with some styling).

aagonzales (approved):

joshblack (approved):

tay1orjones (approved):

PR #2778, opened 2022-03-09
10875 update element docs for v11

dakahn:
Updates the elements docs to match usage guidance for our v11 components 🏄🏾


dakahn:
recheck

PR #2782, opened 2022-03-11
Kingtraceyj patch 1

kingtraceyj:
**Changed**
Updated typography token in style section for body text from `body-compact-01` to `body-01` to match what we are doing in toggletip.


aagonzales:
@kingtraceyj seems like we've got some issues here with a lot of other stuff coming into this PR. Maybe try closing this PR and try opening a clean one again?

kingtraceyj:
yikes. will do.

PR #2792, opened 2022-03-17
docs(website): adds some usage documentation for tabs

abbeyhrt:
REF [#11004](https://github.com/carbon-design-system/carbon/pull/11004)

Updates tabs `code` tab with some more information about v11 changes. Also just adds some notes about Tabs guidance that isn't implemented in code yet. Let me know if that was overkill! I might have over done it. More details about the changes made to tabs are also being added to the storybook docs in the PR linked above

Just an FYI, the code demos won't work on the site until that PR above is merged and released in carbon and carbon is updated here. They aren't _currently_ working so I figured it was okay to go ahead and make the updates so that they should work when the time comes.


aagonzales:
@abbeyhrt I think the v10 vs v11 content would be better suited in the v11 migration guide than on the component page. We want the content of the component pages to have a longevity passed the release phase, for example a year from now people might not care/need to know the difference as much. I think you can leave the v11 content in but rephase to be more of this is how it works. 

I've been thinking a lot about how we let people know there have been changes to a component and I think we can add some temporary inline notifications to component pages that have had major v11 updates. Those notification can then link out the the migration guide if needed. 

Also the deployed failed so I can't view the preview.

abbeyhrt:
@aagonzales gotcha, that makes sense to me. I really like the idea of a temporary inline notification for components that have changed in v11. The main reason I put the information where I did was to have the changes called out in the context of the Tabs guidance and an inline notification would totally accomplish that for me. I'll update the guidance to be less v10 vs v11 specific!

It looks like the deploy previews failed for something unrelated to my PR, hopefully it's a fluke and redeploying will fix it.  🤞

dakahn (approved):

src/pages/components/tabs/usage.mdx
```diff
@@ -187,6 +177,7 @@ B. Indicator <br /> C. Icon
 
 </Column>
 </Row>
+<Caption>Note: The icon to the left of the tab  and the secondary label is an example of a pattern. It is not currently implemented in the component.</Caption>
```
aagonzales, line 180: I thought these were the features that are being added in v11? 

That is what has been in the migration docs for several months now. FYI @kingtraceyj
https://v11.carbondesignsystem.com/migrating/guide/design#tabs

Are these features going to be added soon to the carbon components or are they never going to be added?

tay1orjones (approved):

src/pages/components/tabs/usage.mdx
tay1orjones: Yeah, echoing the similar [convo](https://ibm-studios.slack.com/archives/GSCEH4RSP/p1647981403231799) we had, I do agree we could add some clarity to the language here.

The API has been overhauled to be much more composable to enable these types of configurations. We're no longer restricting what can be put inside a tab.

For now, a few of the new tab modifiers will be something the user can configure, and we will continue to deliver more for it post-GA to provide a better experience out of the box. A style recipe or layout helper, something along those lines so users don't have to configure it all on their own.

tay1orjones (changes requested):
Just want to pop a "request changes" on here to clarify the language around the composable tab features

src/pages/components/tabs/usage.mdx
tay1orjones: Just taking a shot in the dark here, something to iterate on:

```suggestion
<Caption>Note: The optional secondary label and icon are patterns enabled through the composable component API. This requires additional styling and configuration.</Caption>
```

src/pages/components/tabs/usage.mdx
```diff
@@ -156,6 +145,7 @@ are always at least two tabs and one is selected by default. Icons are optional.
 
 </Column>
 </Row>
+<Caption>Note: The close icon to the left of the tab is an example of a pattern. It is not currently implemented in the component.</Caption>
```
tay1orjones, line 148: Same format here, just an idea

```suggestion
<Caption>Note: The close icon button pattern is enabled through the composable component API. This requires additional styling and configuration.</Caption>
```

aagonzales (changes requested):

src/pages/components/tabs/usage.mdx
```diff
@@ -410,6 +415,12 @@ label.
 </Column>
 </Row>
 
+<InlineNotification>
+  Note: Tabs with an icon is an example of a pattern. It is not currently
+  implemented by default in the component, it is instead enabled through the
+  composable component API. This requires additional styling and configuration.
+</InlineNotification>
```
aagonzales, line 422: Oops I suggested the wrong note to be an inline notifications. This one is just a caption

```suggestion
<Caption>
  Note: Tabs with an icon is an example of a pattern. It is not currently
  implemented by default in the component, it is instead enabled through the
  composable component API. This requires additional styling and configuration.
</Caption>
```

src/pages/components/tabs/usage.mdx
```diff
@@ -276,6 +278,9 @@ to drive visual rhythm by spacing content in multiples of two columns and
 aligning the beginning and ending of the tab elements with content below the
 tabs when possible.
 
+_Note: This is not currently implemented in the components, updates for
+alignment to the grid are incoming!_
```
aagonzales, line 282:
```suggestion
<InlineNotification> Note: This is not currently implemented in the components, updates for
alignment to the grid are incoming! </InlineNotification>
```

aagonzales (changes requested):
Looks good, except for the mistake I made in my previous suggestion. I inlined the wrong comment.

tay1orjones (approved):

aagonzales (approved):

PR #2804, opened 2022-03-22
docs(Tooltip): update with v11 components

dakahn:
Part of https://github.com/carbon-design-system/carbon/issues/10889


tay1orjones:
@dakahn Hey I pushed an update to this, I think you just needed it to be the `filename.module.scss` instead of `filename.scss` so it would get picked up appropriately by Gatsby.

Unrelated to this I think there's something up with the markup for the live demo, it's giving an error about how something "expected to receive a single React element child" 🤔

aagonzales (commented):
I'm not seeing anything render in the live demo at all.

src/styles/_page.scss
```diff
@@ -90,3 +90,22 @@ section[role='tabpanel'] [class*='--col'] + [class*='--col'] {
 .tab--columns_label {
   margin-bottom: 0.5rem;
 }
+
+//---------------------------------------
+// Tooltip
+//---------------------------------------
+
+.tooltip-trigger {
+  @include button-reset.reset();
```
jnm2377, line 99: looks like it's failing bc of this. do we need to include it at the top? @tay1orjones

tay1orjones: I think so, yes. `*.module.scss` files are provided a context already containing the utility modules like button, theme, convert, etc.

`.scss` files without `module` are not provided that context and instead need to `@use` the modules needed.

dakahn:
@aagonzales super sorry you got tagged on this too soon. this is a draft PR as it stands. my apologies

src/styles/_page.scss
dakahn: maybe better i move this to a module then -- can do 👍🏾

Issue #2785, opened 2022-03-11
Fix v11 website issues


tay1orjones:
From #2764, a number of minor issues were found that need to be addressed in follow-up PRs

- [x] Some chart styles are not being applied because charts is using the `bx` prefix instead of `cds`
- [x] The grid implementation demo page is broken (will be fixed by upcoming PR https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1233
- [x] the type `object Module` [issue](https://github.com/carbon-design-system/carbon-website/pull/2764#issuecomment-1064348437) has a fix delivered in `main` that needs ported to `next`

- [x] Feedback dialog styles are off. But this may be something we missed in GTC. I'm unsure.

<img width="395" alt="Screen Shot 2022-03-10 at 1 47 47 PM" src="https://user-images.githubusercontent.com/32556167/157911527-8e584d51-ba93-4379-9dc4-4ad9898222a8.png">

(how it should be)
<img width="417" alt="Screen Shot 2022-03-10 at 1 48 07 PM" src="https://user-images.githubusercontent.com/32556167/157912306-f005b97d-5ad5-4654-82ee-f22952d89848.png">

- [x] In the homepage, the feedback dialog can't be opened at all, and it's opacity is less than 1 for some reason. 

<img width="107" alt="Screen Shot 2022-03-10 at 1 48 39 PM" src="https://user-images.githubusercontent.com/32556167/157911539-4e63f1cf-f236-4c7d-9a36-c9d219455931.png">

- [x] some of the codesnippets that have file structures are no longer aligned. This feels nitpicky, sorry. But they're working correctly in the current site, so just wanna make sure we capture that somewhere to fix either now or later.

<img width="892" alt="Screen Shot 2022-03-10 at 2 40 21 PM" src="https://user-images.githubusercontent.com/32556167/157911871-9fa45e20-092a-4ae5-94a5-ea88a314a7c3.png">

<img width="859" alt="Screen Shot 2022-03-10 at 2 00 09 PM" src="https://user-images.githubusercontent.com/32556167/157911853-bafb17ce-88b3-48f1-8fe8-0be65ffd7a3d.png">

- [x] Icon page and Pictogram page search bar is too small, and the dropdown has a dark line at the bottom now. The search bar sizing makes it so that you can see some of the content when you're scrolling.

[...]

PR #2820, opened 2022-03-28
fix(dropdowns): use correct v11 `size` prop

tay1orjones:
Part of https://github.com/carbon-design-system/carbon-website/issues/2785

#### Changelog

**Changed**

- Update our usage of dropdowns to use the new size values from v11. 

#### To Review

- View component demos, the dropdowns should now be the proper size
- View the icon library page, the dropdown in the search filter bar should now be properly sized.


joshblack (approved):

tay1orjones:
I pushed an update, I believe the component demo sizes are correct so I reverted them. Added a style fix for the border between dropdowns on the component demo and the width of the dropdowns should now be correct.

laurenmrice (changes requested):
**Live demo**
- The live demo dropdowns that sit on top of the live demo have increased in width and are too long. They should span the width of the demo window only. They should not span the right panel.
- The live demo dropdowns are also missing a vertical rule between them

To be:
Screen Shot 2022-03-28 at 5 34 40 PM

Current in this PR:
Screen Shot 2022-03-28 at 5 34 26 PM

---------

**Icons (library tab) and Pictograms (library tab)**
- I think there were some overrides set on the dropdowns on both of these pages to match the styling of the Search input that it sits next to. Looks like we need to exclude the $border-strong bottom-border in these scenarios. 

To-be:
Screen Shot 2022-03-28 at 5 41 14 PM


Current in this PR:
Screen Shot 2022-03-28 at 5 41 04 PM

jnm2377:
The color palette dropdown is also shorter than the content switcher now

Screen Shot 2022-03-29 at 10 08 45 PM

laurenmrice (approved):
✅

jnm2377 (approved):
unrelated - but as soon as I opened the deploy preview I was like WOAHHH the homepage looks so nice with the light banner updates!

PR #2823, opened 2022-03-29
chore(deps): update to 10.56.0 Carbon and latest gatsby theme

tay1orjones:
This fixes a number of bugs for https://github.com/carbon-design-system/carbon-website/issues/2785

This updates the website to Carbon v10.56.0 in preparation for v11.

It also updates the gatsby theme version to the latest prerelease version.

#### Changelog

**Changed**

- unprefix tab exports
- properly re-emit grid for demo page with higher specificity selectors to override styling from the theme
- updated deps


abbeyhrt:
@tay1orjones this might not be the PR to do it in, but I noticed that `@carbon/react` isn't a dependency for this repo, it's being brought in through other dependencies. Would we want to add it to the `package.json`?

abbeyhrt (approved):

dakahn (approved):

aagonzales (approved):

tay1orjones:
@abbeyhrt yeah, good eye. It's not technically necessary and can open us up to odd interop bugs where the theme uses a version lower than what we specify on the website. That said I dislike having to cut a release in `gatsby-theme-carbon` every time a new carbon version comes out.

I've added it, didn't even change the lockfile 😅 

I've also put in a small update to gatsby theme version that should make sure the overlay buttons for the feedback dialog and "back to top" render on top of all other elements as intended.

PR #2831, opened 2022-03-29
Update usage.mdx

mbgower:
Removed screen reader info; usual editorial tweaks with Keyboard. Made "group label" the consistent name by removing "heading" (as discussed).
Most important change to scrutinize: the repositioning of the 2nd bullet of the label to the group label area. Took a guess at what was meant; may not be correct.
The other change to confirm: I moved the last bullet of Group labels to Checkbox labels and modified one word. It otherwise made no sense to me.


laurenmrice:
@mbgower Maybe this should be a discussion topic. Group labels should be concise and to the point. For other types of form inputs, after selecting something, you have the option to set it as a warning state in case you need to communicate more information based off of a current selection which could be something similar to what you are talking about? Checkbox does not currently have a warning state though.

mbgower:
@laurenmrice I've removed the 'instruction' text, and added it to the next agenda.

With that done, I think this is ready to go?

laurenmrice (approved):

joshblack:
bump @aledavila or @dakahn when you get a sec!

aledavila (approved):

PR #2837, opened 2022-03-30
Update usage.mdx

mbgower:
Added the fact that the inline code snippet is in the tab order, and can be copied by pressing Enter/Space. There IS equivalent written about it Mouse ("An inline code snippet can be copied by clicking anywhere on the snippet itself.") , so it should be here IMO.


dakahn (approved):

abbeyhrt (approved):

laurenmrice (approved):
👍

Issue #2754, opened 2022-02-25
Update color page with v11 tokens


tay1orjones:
**About**

The color page needs to be updated to showcase the new tokens available in v11. This issue is for updating the existing page. 

We'll implement the more broad new design vision in https://github.com/carbon-design-system/carbon-website/issues/2476 in a follow up release post-v11.

PR #2840, opened 2022-03-31
feat: update color page tokens

jnm2377:
Closes https://github.com/carbon-design-system/carbon-website/issues/2754

#### Changelog

- updates color token names
- adds new tokens


aagonzales (approved):
I'm going to do a bit of a partial approval to get this in for tomorrow (today?). The token names and values looked good from a high level pass. Some of the Roles need some updates but we can do that in another PR later.

PR #2861, opened 2022-04-05
docs(vanilla): update vanilla js docs

### LINKED ISSUE #2852: Update vanilla usage page
Pointed out in https://github.com/carbon-design-system/carbon/issues/11129 

For the https://www.carbondesignsystem.com/developing/frameworks/vanilla/ page, we need to update it to reflect v11 usage.

At a high level, this includes updating

- Vanilla package simply re-exports `@carbon/styles` scss assets
  - A minified css bundle is not included, you'll need to add sass compiliation to your project's build to use v11.
- "If you're looking for v10 docs", link to the v10 site
- Remove the references to unpkg cdn, they no longer apply
- ...any other updates that no longer make sense from v10 to v11

aledavila:
Closes #2852
Closes https://github.com/carbon-design-system/carbon/issues/11129

Update Vanilla JS docs to represent v11 usage. 

#### Changelog

**Changed**

- added a link to v10 documentation and changes scss usage now pointing to the scss migration guide. 

**Removed**

- removed cdn mentions

Please let me know if this makes sense or what i need to add/remove


src/pages/developing/frameworks/vanilla.mdx
```diff
@@ -22,6 +22,9 @@ HTML and SCSS partials to build websites and user interfaces. Adopting the
 library enables developers to use consistent markup, styles, and behavior in
 prototype and production work.
 
+For v10 documents please visit our
+[v10 documentation](https://v10.carbondesignsystem.com/developing/frameworks/vanilla).
+
```
joshblack, line 27: Have we been using a consistent pattern for these notices? Wasn't sure if this was more appropriate in an inline notification or not.

src/pages/developing/frameworks/vanilla.mdx
```diff
@@ -71,10 +72,8 @@ carbon-components/
 ├── html
 ├── css
 │   ├── carbon-components.css
-│   └── carbon-components.min.css
```
joshblack, line 74: I think the what's included section likely needs to be updated, same with the install step above.

joshblack (commented):
Love the first pass! I think there might be other sections on here to get rid of, but I'm curious what you're thinking here.

It seems like we could:

1. Remove all JS-related content and layout stuff, talk only about styles

But this isn't super helpful for people, they'll want to use it with a framework to get behavior

2. Leave this page in but clearly mark it as for v10 and no longer supported? (and move it as the last tab)

Curious what you all think here, I'm not super sure one way or another

src/pages/developing/frameworks/vanilla.mdx
aledavila: Oh yes good Idea I’m not sure if there’s consistent patterns but I do agree with being inline notification

tay1orjones, line 27: Could this instead be in an `<InlineNotification>`? The other framework pages are being updated in #2866 to include a blurb about v11 support in an InlineNotification

tay1orjones: hah, I didn't see Josh's feedback above. We can use #2866 as a template for the InlineNotification though

aledavila:
@joshblack thanks for taking a look at this. From my perspective I think we should highlight this as v10 otherwise it gives the perspective that we are still supporting this. This might be a good opportunity to sunset it. But still show how to integrate with the new styles if they need them.

aledavila:
@joshblack @tay1orjones  i updated it now. I left just scss and the codepen and reordered the tabs to put vanilla at the end

tay1orjones (approved):

joshblack (approved):

PR #2863, opened 2022-04-06
chore: add redirects

### LINKED ISSUE #2850: Add redirects
A quick google search will show the first 2 hits are now a 404 page

Image from iOS

Needs to be edited accordingly 👇 
https://github.com/carbon-design-system/carbon-website/blob/main/conf.d/rewrite.conf

aledavila:
Closes #2850

Added new redirects that happen in mobile. In desktop googling `carbon design system v11` gives you the correct search results.

In duck duck go `carbon design system v11` the first results is `v11.carbondesignsystem.com` is this something i can write in this page ? Or is there a redirect in the backend ? 

#### Changelog

**New**

- redirect whats happening release to main whats happening page

Make sure redirects are written correctly


tay1orjones:
@aledavila could it be applied globally for everything from `v11.` to the root domain?

```
rewrite https://v11.carbondesignsystem.com/(.*) /\$1 permanent;
```

That feels iffy syntax-wise, but maybe that's the right syntax? I don't know enough to be sure.

joshblack (approved):
Changes look good, would be awesome if Taylor's suggestions work too!

conf.d/rewrite.conf
```diff
@@ -44,3 +44,7 @@ rewrite /tutorial/vue/(.*) /developing/vue-tutorial/\$1 permanent;
 # Data-visualization
 rewrite /data-visualization/basic-charts/ /data-visualization/simple-charts/ permanent;
 rewrite /data-visualization/advanced-charts/ /data-visualization/complex-charts/ permanent;
+
+#v11
+rewrite https://v11.carbondesignsystem.com/migrating/guide/\$1 /migrating/guide/\$1 permanent;
```
joshblack, line 49: Quick question, would this also work?

```suggestion
rewrite /migrating/guide/\$1 /migrating/guide/\$1 permanent;
```

sstrubberg (approved):

conf.d/rewrite.conf
aledavila: I’m actually not even sure how to text it. Since v11 is deployed. I’m trying to redirect from v11.carbondesignsystem.com to regular carbondesignsystem.com

aledavila:
@tay1orjones actually I was just thinking now. can I redirect a whole domain to regular carbondesignsystem.com through this file? I feel like I’m incorrect.

sstrubberg:
Not to complicate things, but I just ran into [this](https://github.ibm.com/tec-dach/tec-dach.github.ibm.com/blob/develop/gatsby-node.js).

tay1orjones:
@sstrubberg I think that is using (`gatsby-plugin-client-side-redirect`)[https://www.gatsbyjs.com/plugins/gatsby-plugin-client-side-redirect/], which is a client side redirect using `window.location.href`. 

From what I've read, the google crawler prefers server initiated redirects like what we'll have here through nginx

sstrubberg:
> @sstrubberg I think that is using [`gatsby-plugin-client-side-redirect`](https://www.gatsbyjs.com/plugins/gatsby-plugin-client-side-redirect/), which is a client side redirect using `window.location.href`.
> 
> From what I've read, the google crawler prefers server initiated redirects like what we'll have here through nginx

Gotcha! I'm guessing they're using this on the PAL because they're not using nginx when they deploy

tay1orjones (approved):
@aledavila we can always merge and see. Or try a specific subset of pages first and merge, see how it works, then roll it out for a broader set of pages.

PR #2880, opened 2022-04-11
fix(livedemo): add left border

tay1orjones:
This updates all live demo preview containers to now have a left border to avoid the issue described in https://github.com/carbon-design-system/carbon/issues/10889#issuecomment-1077992681

#### Changelog

**Changed**

- add back the left border to the live demo preview container

### DIFF
--- a/src/components/ComponentDemo/ComponentDemo.module.scss
+++ b/src/components/ComponentDemo/ComponentDemo.module.scss
@@ -281,6 +281,7 @@ fieldset.form-group:last-of-type {
   overflow: auto;
   padding: $spacing-05;
   border: $spacing-03 solid $background;
+  border-left: 1px solid $ui-03;
   background-color: var(--cds-background);
   width: 100%;
   display: flex;


joshblack (approved):

aagonzales (changes requested):
Its doing a weird tapper at the corner. Can we get those two border ends to nicely meet

image

aagonzales:
Also I think we probably need to border to go all the way across the top as well. Otherwise there's a gap there with the Gray 10 theme

image

aagonzales:
@tay1orjones so I thought this was a intentional design decision at some point but maybe was just bug. Instead of the line board, these demos used to have this 8px white boarder. I talked to the other designers and seems like this change wasn't intentional. Can we actually just bring that back? 

Its on the [v10 demos](https://v10.carbondesignsystem.com/components/accordion/usage/#live-demo)

tay1orjones:
@aagonzales Ah! That explains a ton 😅 I see now that we updated the border to use the new tokens, so it's no longer a fixed value but changes with the theme. I'll push an update to fix.

aagonzales (approved):
Nice!

PR #2886, opened 2022-04-14
File uploader accessibility

mbgower:
Rewrite of Accessibility information for File Uploader, as discussed on CAG call.


mbgower:
@dakahn I got an email you'd made a comment to me on this, but it looks like it's gone, so I assume you answered your own question?
Going forward, I think it makes sense to create the Accessibility tab content as part of the component work. We may well surface considerations that can help inform or even tweak the design before it is released.

joshblack (approved):

aledavila (approved):

src/pages/components/file-uploader/accessibility.mdx
```diff
@@ -19,82 +19,51 @@ import {
 
 <PageDescription>
 
-The file uploader React Carbon component has been tested against the latest
-[W3C Web Content Accessibility Guidelines (WCAG) 2.1 Level A and AA success criteria](https://www.w3.org/TR/WCAG21/)
-and violations have been identified as high priority issues. This document will
-be updated when these accessibility issues are resolved.
+No accessibility annotations are needed for file uploaders, but keep these considerations in mind if you’re modifying Carbon or creating a custom component.
```
laurenmrice, line 22: > No accessibility annotations are needed for breadcrumbs, but keep these considerations in mind if you are modifying Carbon or creating a custom component.

- I noticed you have a contraction for "you’re" on the File uploader a11y tab in this PR and on the Slider tab but not on the Breadcrumb tab and Code snippet tab. I would just keep it consistent across the board for this same sentence in all places it shows up.

src/pages/components/file-uploader/accessibility.mdx
```diff
@@ -19,82 +19,51 @@ import {
...
-### Automated test
+## What Carbon provides
+
+Carbon bakes keyboard operation into its components, improving the experience of blind users and others who operate via the keyboard. Carbon incorporates many other accessibility considerations, some of which are described below.
+
+### Keyboard interactions
+
+Both variants of the file uploader provide buttons for uploading and removing files.
+The drop target “Drag and drop files here..." also provides conventional button
+keyboard interaction (`Tab` to reach; `Enter` or `Space` to activate). Once a file
```
laurenmrice, line 39: >Once a file has been added, it can be removed by activating the delete (“x”) button after each file name.

- In the Anatomy section on the Usage tab, we describe this as the Delete `x` icon. And on the Style tab we describe it as the Delete icon. The x on the usage tab is styled as code and we say icon instead of button. I think we should align on our terminology for it.

src/pages/components/file-uploader/accessibility.mdx
```diff
@@ -19,82 +19,51 @@ import {
...
+<Caption>Uploaded files can be removed by tabbing to each “x” button and activating.</Caption>
+
+</Column>
+</Row>
+
+## Development considerations
+
+Keep these considerations in mind if you’re modifying Carbon or creating a custom component.
+
+* the 'Drag files' area is constructed as a button to support keyboard operation
```
laurenmrice, line 67: I would capitalize the first letter and end with a period for each line here like you have done on other accessibility tab pages.

mbgower: done

src/pages/components/file-uploader/accessibility.mdx
mbgower: I agree Carbon is all over the place in how this is discussed -- even within the Usage content it varies. My concern with calling this an 'icon' in the context of the keyboard accessibility is that it isn't operabler because it's an icon (which isn't necessarily a control) but because it's a button. My concern with capitalizing "delete" is that that is not the visible label of the button, which is what I think traditionally is capitalized. This is an odd button, but this helps from IBM Style:
> When you write about an item that is displayed in a graphical user interface, match the capitalization and spelling of the item in your writing. For example, if a label is displayed with initial uppercase letters or all uppercase letters, write the label the same way in the documentation

laurenmrice: Oh sorry, I wasn't concerned about capitalization, just saying icon versus button. If you think that using button for keyboard accessibility context makes more sense, than we can keep it as is.

src/pages/components/file-uploader/accessibility.mdx
mbgower: Nice catch. I think Anna asked for the contractions to be removed in some of the reviews. I did point out to her that Carbon uses the "you're" contraction all over the place (57 occurrences in 35 files), so I think we can just make a decision and stick with it.
Do you have a preference?

mbgower: made change across accessibility content I've updated.

laurenmrice (approved):
👍

PR #2917, opened 2022-05-03
[DO NOT MERGE] docs(tutorial): update step 1 & 2 to v11 instructions

jnm2377:
Closes https://github.com/carbon-design-system/carbon/issues/10991
rel: https://github.com/carbon-design-system/carbon-website/pull/2912

DO NOT MERGE UNTIL STEPS 3 & 4 HAVE BEEN APPROVED AS WELL 
will merge PRs on the same day to make sure nothing is out of sync

#### Reviewing

It would be most helpful if we had reviewers walk thru the steps (1 & 2) to make sure we didn't miss anything in the instructions

Note - the branches for v11 are completely new branches than v10. But all should be clear in the instructions.


src/pages/developing/react-tutorial/step-2.mdx
```diff
@@ -262,73 +240,76 @@ import {
...
-      </Tab>
-    </Tabs>
-  </div>
-</div>
-```
+The `<TabList>` is the tab navigation, and the `<TabPanels>` contains each tab
+content. You'll notice that within each `<TabPanel>`, we nested another `<Grid>`
+and `<Column>`. This is because the parent column
+`<Column lg={16} md={8} sm={4} className="landing-page__r2">` is acting as a
+feaux row, and each tab within that row contains the actual column content. When
```
aagonzales, line 304: Spelling error 
```suggestion
faux row, and each tab within that row contains the actual column content. When
```

jnm2377:
> Looks good from design expect one typo and a broken link.
> 
> The link to the type page is broken under Step 2 > Style landing page > Banner. It wasn't in the suggested area to fix.
> 
> Also side note: does the Sketch file associated with the tutorial need to be updated as well? I noticed one of the type tokens being used was still showing v10 ($productive-heading-03).

@aagonzales Fixed the typo. 👍🏽 

I am unsure about any of the design assets or images. I suppose it would make sense to make sure the sketch file is using v11. It hadn't really occurred to me. We were just migrating the code and instructions, but it sounds like we might need a designer to migrate design assets?

aagonzales:
Yeah I see the preview example but I didn't actually see a Sketch file download anywhere. And then there's this image oh sketch as well which is showing an old token name.

image

PR #2923, opened 2022-05-05
fix(website): update component demo for dropdown

abbeyhrt:
Closes [#11284
](https://github.com/carbon-design-system/carbon/issues/11284)
{{short description}}

Adds docgen info for FilterableMultiselect. Removes `light` prop and adds `warn`, `size`, and `invalid`.

Also updates docgen info for other list box components so that the size prop works in the knobs as expected


joshblack:
@abbeyhrt should this be updated in `usage.mdx`, as well, or is that out of scope for this PR?

tay1orjones (approved):
We could always put the usage page updates in a new PR 👍

joshblack:
@abbeyhrt seems like there is a merge conflict and then should be good to go!

laurenmrice (commented):
I notice that when switching to the warn or invalid props, the input moves slightly. Is this an issue with react or with the live demo?

dropdown

abbeyhrt:
@laurenmrice It should be a problem with the demo, I can look into it to get it to stabilize!

laurenmrice (approved):

PR #2930, opened 2022-05-09
fix(component-demo): update knobs to work with defaultValue

joshblack:
This PR updates our Knobs to support a broader range of values from the react-docgen.json file. This should work with some of the new component demoes we are looking to add.

#### Changelog

**New**

**Changed**

- Update Knobs to work with `defaultValue` of various types, both string and boolean

**Removed**


jnm2377 (approved):

aagonzales (approved):
I wasn't entirely sure how to test this but I trust you.

PR #2953, opened 2022-05-23
chore(website): add common.js script

abbeyhrt:
Closes https://github.com/carbon-design-system/carbon/issues/11460

Removed fathom script and adds common script. I added what made sense from [this](https://github.com/carbon-design-system/design-language-website/blob/master/gatsby-ssr.js) but let me know if something is missing!


sstrubberg:
Can we add this without removing Fathom? I'd like to remove Fathom at the same time that we add Google Analytics.

abbeyhrt:
@sstrubberg yeah for sure! I'll make that update

sstrubberg:
There's also [this piece](https://github.com/carbon-design-system/design-language-website/blob/075fb28bfdaa28b4c663669935ad3578eb1bc773/gatsby-ssr.js#L18-L21).

How do we know if this is working?

sstrubberg (changes requested):
Got this from @jeffchew. 

https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/blob/main/src/gatsby-theme-carbon/components/Footer/index.js#L59-L68

Instead of the cookie preferences in the lower-right-hand corner *not* using Plex, we could just add it to the footer.

image

abbeyhrt:
@sstrubberg What would I change the owner and the site ID to? Linda was saying something in round robin about changing the owner(I think) and obviously the site id has to change to be unique. 
```
          siteID: 'DESIGN_LANGUAGE',
          country: 'US',
          industry: 'Design',
          owner: 'Mike J Abbink/New York/IBM',
```

abbeyhrt:
@sstrubberg I made the changes to the site data but adding this into the footer is a bigger ask since we don't currently shadow the footer component. It's definitely doable but isn't a quick task and I feel like that would need to be a different work item/issue since it feels out of scope of just adding common.js to the website. If we're talking sizing, I would originally have thought of this issue as a 1 and but with adding a new shadowed footer component, it brings it up to like a 5, at least to me and I would prefer for it to be a separate issue/PR

aledavila (approved):
lgtm

sstrubberg:
@abbeyhrt when i look at the deploy preview, I'm not seeing the cookie preference button anymore. Did something change?

jeffchew:
@sstrubberg I still see `ibm-common.js` loading (and all of the other tags that come with it), so I think it's fine. I sometimes see this behavior too, but don't think it's on our end. Also, a ton of the tags are blocked (including cookie preferences) if you have an adblocker enabled.

sstrubberg (approved):

PR #2977, opened 2022-06-06
Update usage.mdx

mbgower:
I neglected to turn this commit into a PR when I created it back in March. The contents were reviewed on CAG call back then.


mbgower:
Please note, as per my comments in the [review](https://ibm.box.com/s/8snlzgw10usjbdln7d2g790iqanzkpf6), that the image under Overflow content needs to be updated. It incorrectly shows no focus on the first item in the menu. When a user opens a menu, the focus needs to be placed on the first item in the submenu.
I also believe that it should show breadcrumbs 1 and 2 preceding the overflow ellipse, not just one? That is how it is implemented in Carbon -- although it is possible to read this text to mean either "the first two and the last two page links should be shown..." OR "the first page link and last two page links should be shown" 
image

jnm2377 (approved):
Hey @mbgower looks like you just need to run `yarn format` and commit those changes to get ci to pass. Otherwise lgtm. 👍🏽

dakahn (approved):

Issue #2987, opened 2022-06-13
Site does not have working `og:image` meta tag


tay1orjones:
## Detailed description

As pointed out in [slack](https://ibm-studios.slack.com/archives/C03468NQVLJ/p1655146834478299), the `og:image` tag does not work. It is using a relative path that is not valid. The image isn't in the static folder either if you attempt to correct the path. Using a relative path isn't correct anyway, by what I can tell it's always supposed to be an absolute url.

This could be related to gatsby v3 vs v4 that we updated to as part of the v11 push. It also could be something stemming from gatsby-theme-carbon. I'm not sure.

Until we fix this, link previews or unfurled previews in slack won't display what we prefer
![image](https://user-images.githubusercontent.com/3360588/173448733-c82e4d79-3057-4ac3-85cc-73b8785d023b.png)

PR #3005, opened 2022-06-22
fix(og:image): use absolute path

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/2987

Uses the absolute path to the static GitHub hosted image rather than a relative path. 

#### Changelog

**Changed**

- changed `og:image` URL to an absolute path rather then relative


tw15egan:
@tay1orjones any idea why this build is failing?

abbeyhrt (approved):

PR #3022, opened 2022-06-29
corrected link to toggletip

mbgower:
The link was 404ing since I treated "toggle tip" as two words, and so put a dash in the link

Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}


laurenmrice (approved):

aledavila:
I think it might be better to remove the link all together for now as Im still get a 404 page since as Taylor said. We don’t actually have an accessibility page atm. @mbgower

PR #3024, opened 2022-06-29
Added Bluetalks to the Migration page

sstrubberg:
Just adds a ResourceCard for Josefina's talk


aledavila (approved):
Lgtm

aagonzales (approved):
Looks good, but I wonder if we should rename the section to something other than "YourLearning" maybe like "Release playbacks"?

PR #3061, opened 2022-07-25
Toggle accessibility

mbgower:
Updating accessibility content and images for toggle


aagonzales (commented):
The do/dont source images should not include the indicator icons. They are provided in code. Source images need to be updated to correct icon overlap
image


In the redundant state information image, I'd also recommend adding a divider line between  the two examples

mbgower:
@aagonzales totally agree, and had already pointed those things out to @laurenmrice in a direct slack on this PR, since she normally revamps the images as part of the PR QA process

mbgower:
@aagonzales I forgot she had told me she was on holiday and asked me to flag you, so here's what I sent her.
I can ask Jess to update for now, if you don't have cycles.

>  The do/don’t image pair seems to apply its own styling, so I take it the images don’t need that. I assumed you would address when you clean up — or maybe you want me to NOT code as a do/don’t pair?
Incidentally, I think the second image might work better as a Do pair, as opposed to one image?

mbgower:
@aagonzales I have pushed new images (from Jess) along with modified captions and alt text

aagonzales (approved):
looks good now!

PR #3064, opened 2022-07-28
Inline loading accessibility

mbgower:
Accessibility content update for inline loading, both text and images


src/pages/components/inline-loading/accessibility.mdx
```diff
@@ -14,197 +14,66 @@ import {
...
-## Accessibility considerations
+### Keyboard interaction
 
-This component has been validated to meet the
-[WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) and
-[Section 508](http://www.section508.gov/) accessibility guidelines, however
-changes made by the content owner can affect accessibility compliance. Be sure
-to review and follow the guidance in this section when updating or adding new
-content to this component.
+The inline loading component is not itself keyboard operable; however, when inline loading is in an active state, it may temporarily replace or disable an associated component. For instance, if inline loading is showing the result of a button action, relevant input controls as well as other actions may be disabled (and so be inoperable) until the loading state changes to either error or complete.
```
abbeyhrt, line 38: The button switching to disabled when inline loading is showing isn't baked into the component. We have a story showing how a user could code that up themselves but it isn't a feature of the component itself.

 I'm not sure how to update this guidance to reflect that. I just worry that people will think that it is inline loading that is disabling relevant input controls when that would need to be implemented separately by users

abbeyhrt: Maybe something like, "We recommend disabling relevant input controls when inline loading is showing due to the result of a button action. To see an example of how to do this, you can check out our storybook [here](https://carbon-components-react.netlify.app/?path=/story/components-inlineloading--ux-example)"?

abbeyhrt (approved):
I made a note about the keyboard interaction but otherwise this all looks good to me! 🎉

abbeyhrt:
Just FYI, I ran format to get all of the checks to pass.

src/pages/components/inline-loading/accessibility.mdx
mbgower: Abbey, I assume you're talking about this:
> The inline loading component is not itself keyboard operable; however, when inline loading is in an active state, it may temporarily replace or disable an associated component. For instance, if inline loading is showing the result of a button action, relevant input controls as well as other actions may be disabled (and so be inoperable) until the loading state changes to either error or complete.

I think the use of "may" here conveys what's going on. Remember, we're just trying to capture the effects on keyboard operation -- the gist of this being that if something is disabled, obviously it can't be keyboard accessible. I'll review the language and see if it can't be made clearer that this disabling function is NOT part of Carbon.

mbgower: How about "... however, when inline loading is in an active state, it can be implemented to it temporarily..."

> The inline loading component is not itself keyboard operable; however, when inline loading is in an active state, it can be implemented so that it temporarily replaces or disables an associated component. For instance, if inline loading is showing the result of a button action, relevant input controls as well as other actions may be disabled (and so be inoperable) until the loading state changes to either error or complete.

laurenmrice:
I have got some image edits to this PR. I will create another PR to make those changes and link here for reference.

mbgower:
@laurenmrice can you take a look at my suggested wording above to address Abbey's concern, and incorporate into yours if that works for you?

Issue #3045, opened 2022-07-14
Date picker Usage and style information needs to be addressed


mbgower:
This issue covers material additional to the two issues specifically opened up for Date picker https://github.com/carbon-design-system/carbon/issues/11805 and https://github.com/carbon-design-system/carbon/issues/11800

A review of the Usage and Style pages was carried out as part of an accessibility review. A number of design and implementation considerations for the Date picker component have been flagged in the [Date picker usage.pdf](https://ibm.box.com/s/d863av8pyed8pj7mwi3ck8oo1npf3t1m) and [Date picker style.pdf](https://ibm.box.com/s/69gitja8716wjikuwmmh6ufhyvme0c23) documents.

These should be opened in Adobe Acrobat Reader (not in a browser) so that all the comments, replies and markings on status can be reviewed. (Any comments that have been addressed through updates to the Accessibility tab are marked as Complete or Resolved).

There are over 50 comments in the Usage document alone, attesting to the variety of considerations necessary for this component, among them:

aagonzales:
My notes on actionable task from Gowers PDFs

## Kits
- Add format by default in helper text (in addition to placeholder or remove placeholder?)


## Website 

**Content edits**
- Images
   -  Update all images with correct helper text placement.
   - Website: clarify variant vs type language (sub-variants are types)
- Intro paragraph:
    -  break apart into two sentences, one for date picker and one for time picker. Better clarity on variants. 
- Overview: 
    - The kind of date or time (exact, approximate, memorable) you request from the user will determine which picker (simple, calendar or time picker) is best to use.
    - When to use: break into two bullet points? (Instead of paragraph)
    - Variants/Calendar: say there are two types of calendar variants (single and range)
    - Hyperlink variant names
- Anatomy: 
    - add date format as a number, include in helper text or label area.
    - Day: Add link out to calendar variants for specific day type
    - Add weekday call out
- Sizes: calcify the sizes are for the input only. Calendar menu remains is a fixed size and does not change with input size.  (Example: show with calendar? Or maybe this is in the calendar variant section).
- Universal behaviors: 
    - Add warning state
- Simple date: when to use: when days of the week are irrelevant 
- Time picker: add time format 24hr vs 12hr and examples
- Modifiers
    - Remove light variant
- Related 
    - Using ind inputs: remove mention of time input, just for date

Style tab
 - Color: add “Day: start range”
 - Typography: add helper text, warning message, and placeholder

**Page structure**
[...]

PR #3096, opened 2022-08-16
docs(date-picker): gower review

aagonzales:
Closes https://github.com/carbon-design-system/carbon-website/issues/3045

Recommended immediate updates from @mbgower's review of the date picker docs. Some updates in Mike's original PDF like page structure will be addressed in a future issue. 

[See issue](https://github.com/carbon-design-system/carbon-website/issues/3045#issuecomment-1204289674) for list of changes applied in this PR.


dakahn (approved):

src/pages/components/date-picker/usage.mdx
```diff
@@ -30,23 +30,23 @@ times.
...
+Date pickers allow users to select past, present, or future dates. The kind of
+date you are requesting from the user will determine which date picker (simple
+or calendar) is best to use. A time picker helps users select a specific time.
+You can customize the time picker's format depending on location or need.
 
 ### When to use
 
-Use date and time pickers when you are asking the user for a time or date, or
-for scheduling tasks.
+- When asking the user for an exact, approximate, or memorable date or time
```
laurenmrice, line 40:
```suggestion
- When asking the user for an exact, approximate, or memorable date or time.
```

src/pages/components/date-picker/usage.mdx
```diff
@@ -30,23 +30,23 @@ times.
...
+- For scheduling tasks
```
laurenmrice, line 41:
```suggestion
- For scheduling tasks.
```

laurenmrice (changes requested):
- On the Style tab under the Structure section for "Time picker", the end of the description sentence has a typo of `Hi` that should be removed.
Screen Shot 2022-08-17 at 10 50 27 AM

- When you show the open calendar menu in images, the previous/next month days should be using `$text-secondary` color token. I think this is a bug in the symbol itself and it just needs to be updated. Might be better to update those images in a later PR once the bug is fixed in the symbol so we don't have to break them all?

Screen Shot 2022-08-17 at 10 54 15 AM

aagonzales:
Oh no, that's a problem with the library component. I'll have to push a change from there first to update the calendars. Good catch though

tay1orjones (approved):

laurenmrice (approved):
🎉

PR #3118, opened 2022-08-26
Update color-tokens.js

aagonzales:
A few color tokens bugs across the themes on the color table have been pointed out

- Support colors are out of order

Gray 10
- Background-active
- Layer-accent-active-01
- Highlight 

Gray 100
- Background states formatting
- Field-hover-02
- Text on color disabled

G90
- Field-hover-02
- Text on color disabledGray 10
- Background-active
- Layer-accent-active-01
- Highlight


src/data/guidelines/color-tokens.js
```diff
@@ -1240,12 +1240,12 @@ const colorTokens = {
           hex: '#8d8d8d',
         },
         g90: {
-          name: 'Gray 50',
-          hex: '#8d8d8d',
+          name: 'Gray 10, 25% opacity',
+          hex: '#f4f4f4 @ 25%',
```
tw15egan, line 1244: Seems like this doesn't match the code. Should `text-on-color` be changed to `Gray10` or should we just use `Gray10` instead of referencing `text-on-color` in `text-on-color-disabled`?

Screen Shot 2022-08-30 at 6 20 14 PM


Same issue with the `g100` token

aagonzales: Oh oops good catch. That's correct in code. This should be White here.

aagonzales: I just say the color because its hard to reference other tokens in this context.

tay1orjones (approved):

thyhmdo (approved):

PR #3182, opened 2022-10-01
docs(progress-bar): add component live demo

janhassel:
Adds a component live demo of the `ProgressBar` to its "Code" page and the "Usage" page.

#### Changelog

**New**

- Added ProgressBar component demo

**Changed**

- Updated `react-docgen.json` to include ProgressBar props


tw15egan:
Noticed that the styles don't seem to be loading properly when the status is changed

Screen Shot 2022-10-03 at 1 16 05 PM

janhassel:
@tw15egan That's weird. It works in the gatsby dev build, but I can reproduce this issue in the production build.
Looking at the generated stylesheet, the status style rules are missing even though they are present in the `progress-bar.scss` source from `@carbon/styles` (https://unpkg.com/browse/@carbon/styles@1.13.0/scss/components/progress-bar/_progress-bar.scss, lines 96-126).

The styles were included since the v11.3.0 release: https://github.com/carbon-design-system/carbon/commit/c7521ed5a06e7dd53a0d843471fcc0a31f9b8513#diff-0445797061c6a9508f633c1ac24f253522e6e5a2de18b9ef115ca5b9b6717d5c

Do you have an idea why they might not be picked up (or thrown out) in the gatsby production build?

tw15egan:
@janhassel hmm very strange. Could it possibly be resolving to a previous version of the package?

tw15egan:
Hmm that also didn't seem to work, we might need to update the `gatsby-theme-carbon` package directly

janhassel:
@tw15egan Ah, that makes more sense. Thank you! I noticed `gatsby-theme-carbon` already updated [`@carbon/react` to v1.11.0 a](https://github.com/carbon-design-system/gatsby-theme-carbon/commit/8bece9302c6851dc8f3a7d3ddc2fe78af3719efe) few weeks back but it hasn't been released yet.

@sstrubberg @tay1orjones Any chance you could push a `3.0.0-next.13` release so that we can pull in the newer Carbon version?

tay1orjones:
@janhassel Yes! I just published (`v3.0.0-next.13`)[https://github.com/carbon-design-system/gatsby-theme-carbon/releases/tag/v3.0.0-next.13]

tw15egan:
Thanks, @tay1orjones! Pushed up a commit that updates the package and it seems to be working 🎉

tw15egan (approved):

thyhmdo (approved):

PR #3193, opened 2022-10-11
ci(actions): automatically open content sync issues on platform

tay1orjones:
#### Changelog

**New**

- added action that will open a new issue on the platform repo whenever a PR is merged here on the carbon-website repo that contains changes of `.mdx` files


.github/workflows/platform-content-sync.yml
```diff
@@ -0,0 +1,24 @@
...
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: carbon-design-system/carbon-platform
+          ref: main
+      - name: Create issue on platform repo
+        run: |
+          gh issue create --title "[Content sync]: carbon-website PR ${{ github.event.number }}" --body 'The following pull request on carbon-website was just merged. It contains .mdx content changes that may need synced to platform.
```
mattrosno, line 21:
```suggestion
          gh issue create --title "[Content sync]: carbon-website PR ${{ github.event.number }}" --label "role: dev 🤖" --label "service: web-app 🌎" --body 'The following pull request on carbon-website was just merged. It contains .mdx content changes that may need synced to platform.
```

Can we add two labels by default? I think this is the correct syntax per: https://cli.github.com/manual/gh_issue_create

.github/workflows/platform-content-sync.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  if_merged:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: carbon-design-system/carbon-platform
+          ref: main
```
jdharvey-ibm, line 18: Might not need the checkout/clone here since this doesn't look to hit the repo code at all.

mattrosno: @jdharvey-ibm do you have alternative suggestion?

jdharvey-ibm, line 18:
```suggestion
```

@mattrosno This first job step should just be able to be removed entirely.

jdharvey-ibm: Removing this would I think require the `gh issue create` command below to have a `--repo=carbon-design-system/carbon-platform` flag on the command.

tay1orjones: Yeah exactly. I just did the checkout so `gh` would have the proper context to open the issue on the right repo.

Using just a single command it's easy enough to use the `--repo` flag but didn't know if there were additional commands or things you all might want to add in addition to the issue being opened.

I'm good with either approach 👍

jdharvey-ibm: @tay1orjones  I guess we have lots of available runners and actions minutes, so the time incurred by the checkout is probably not an issue 😅

jdharvey-ibm (approved):

PR #3288, opened 2022-11-28
pin node version at 16

alisonjoseph:
Attempt to fix Vercel deployment issue 
https://vercel.com/carbon-design-system/carbondesignsystem/5jZswT9eJvMeTaJbFAJ6ezfCY1GM


tw15egan (approved):
Looks like that did the trick 👍🏻

PR #3294, opened 2022-12-01
Carbon-Platform: add type="do" to unspecified DoDont

francinelucca:


mattrosno (approved):
I see `type="do"` is a Gatsby theme MDX default, so this should be a no-op for carbon-website, while this fixes Platform MDX as there is no default there.

https://gatsby-theme-carbon.vercel.app/components/DoDontRow#do-and-dont

PR #3303, opened 2022-12-08
Modal accessibility content and images

mbgower:
Rewrite of accessibility content for Modal component


tay1orjones (approved):

tw15egan (approved):

src/pages/components/modal/accessibility.mdx
```diff
@@ -17,400 +17,109 @@ import {
   ListItem,
 } from '@carbon/react';
 
+<PageDescription>
+
+Design annotations are needed for specific instances shown below, but for the
+standard modal dialog component, Carbon already incorporates accessibility.
```
aagonzales, line 23: Edit to add clarity, I had to read the sentence a couple of times to make sure I understood it correctly. Just a suggestion.

```suggestion
Design annotations are needed for specific instances shown below, except for the
standard modal dialog component that Carbon already incorporates accessibility.
```

src/pages/components/modal/accessibility.mdx
```diff
@@ -17,400 +17,109 @@ import {
...
-### iOS screen reader tests
+### Focus handling
+
+When the dialog appears, the first item that gets focus depends on the type of
+dialog. Passive dialogs only contain a close button (X), so that takes focus.
+For dialogs which prompt for confirmation or user decision, the primary button
+takes focus (regardless of number of buttons). For destructive interactions, the
+“cancel” button takes focus, not the red danger/delete button. The tab order
+should proceed left and down from whichever item has focus then wrap back to the
+Close button (X).
```
aagonzales, line 78: Close would not be capitalized as it's not a proper noun nor in an emphasis style as a label. 
```suggestion
close button (X).
```

src/pages/components/modal/accessibility.mdx
```diff
@@ -17,400 +17,109 @@ import {
...
+## Design recommendations
+
+### Designate the input that takes focus in a transactional
+
+The only time the starting focus would not be on a button is where a dialog
+contains input fields. For such transactional dialogs, designers should annotate
+that the first input field should receive focus. Note that text links are not
+considered a user input in this context.
+
+<DoDontRow>
```
aagonzales, line 99: Shouldn't the word focus in the image be capitalized? "Focus..."

image

aagonzales (commented):
It looks good except for a couple of minor content suggestions or questions. Also I noticed you used both the term "dialog" and "modal." Is the use of dialog in some place over modal intentional? This is the modal component so I wonder if we should just stick to modal or be specific and say "modal dialog" instead of just "dialog."

src/pages/components/modal/accessibility.mdx
aagonzales: Also I think the content in the images could be cleaned up and add a little more context like so (my image is just hacked together with a screen shot)

image

src/pages/components/modal/accessibility.mdx
mbgower: @aagonzales That's the stock pre-amble. While I agree it can get improved on, I'm not sure the rewrite improves it enough.
For most components, we're conveying that 'you're all good.' For some, we're trying to convey 1) most of the stuff is taken care of 2) in some (but depending on the component, not all) situations you need to make some design annotations.
I suggest we come back to the pre-amble with fresh eyes after the holidays and do a global update once we get something everyone thinks is an improvement :)

src/pages/components/modal/accessibility.mdx
mbgower: technically it is the accessible label name, but I take your point :)

src/pages/components/modal/accessibility.mdx
mbgower: @laurenmrice I'll let you tackle these image changes. "Focus" can be capitalized if that's preferred; no idea what you folks use for annotation conventions

src/pages/components/modal/accessibility.mdx
aagonzales: Yeah that's fine to come back to this. I think its just the phrasing/sentence structure that is causing confusion for me. It would be great if we could have a content designer review the boilerplate intros and content.

aagonzales (approved):
Alright then i think this is good to go.

PR #3306, opened 2022-12-09
ci: Use gatsby-cache action to cache gatsby build outputs

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

I created an action to cache Gatsby's build outputs. This action is easy to set up. 

**AS-IS**

```yaml
- name: Gatsby Cache Folder
  uses: actions/cache@v3
  id: gatsby-cache-folder
  with:
    path: .cache
    key: ${{ runner.os }}-cache-gatsby
    restore-keys: |
      ${{ runner.os }}-cache-gatsby
- name: Gatsby Public Folder
  uses: actions/cache@v3
  id: gatsby-public-folder
  with:
    path: public/
    key: ${{ runner.os }}-public-gatsby
    restore-keys: |
      ${{ runner.os }}-public-gatsby
- name: Build site
  run: yarn build:clean
```

**TO-BE**

```yaml
- uses: jongwooo/gatsby-cache@v1.2.1
  with:
    key: ${{ runner.os }}-gatsby-build
    restore-keys: |
      ${{ runner.os }}-gatsby-build
- name: Build site
  run: yarn build
```

## References

- [https://github.com/jongwooo/gatsby-cache](https://github.com/jongwooo/gatsby-cache)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,24 +57,13 @@ jobs:
           node-version: 16.x
       - name: Install packages
         run: yarn install --frozen-lockfile --network-timeout 300000
-      - name: Gatsby Cache Folder
-        uses: actions/cache@v3
-        id: gatsby-cache-folder
+      - uses: jongwooo/gatsby-cache@v1.2.1
         with:
-          path: .cache
-          key: ${{ runner.os }}-cache-gatsby
+          key: ${{ runner.os }}-gatsby-build
           restore-keys: |
-            ${{ runner.os }}-cache-gatsby
-      - name: Gatsby Public Folder
-        uses: actions/cache@v3
-        id: gatsby-public-folder
-        with:
-          path: public/
-          key: ${{ runner.os }}-public-gatsby
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


jongwooo:
I have read the DCO document and I hereby sign the DCO.

jongwooo:
@tay1orjones, Thank you for your comment! 

This action supports caching in the same way as `actions/cache`, with less setup. And it internally sets up Gatsby's [conditional page builds](https://www.gatsbyjs.com/docs/reference/release-notes/v3.0/#incremental-builds-in-oss), enabling fast builds. This action also being used at [nodejs.dev](https://github.com/nodejs/nodejs.dev/blob/da95e60a58256e17a067a2aba2a4fff4e8ca8641/.github/workflows/github-pages.yml#L33).

tay1orjones:
My understanding is that we already have conditional page builds - Gatsby v3 turned them on by default as long as both `.cache` and `/public` are cached.

Just to be totally candid with you, I don't see a strong value proposition for adding this action. Also I would prefer to not introduce another dependency in our pipeline.

PR #3324, opened 2023-01-02
typo fix

jeanservaas:
removed unnecessary parenthesis

see combo box description

image

PR #3326, opened 2023-01-03
fix(form-patterns): ran yarn format

sstrubberg:
Addresses Jeannie's prettier error on https://github.com/carbon-design-system/carbon-website/pull/3324

Let me know if there's an easier way to do this.


alisonjoseph:
LGTM! I think in the future you can just edit her branch directly vs. opening a PR, but this totally works to.

alisonjoseph (approved):

Issue #3207, opened 2022-10-18
Fluid inputs: FluidNumberInput docs


tay1orjones:
Add to website branch to house all fluid input updates (do not merge)

https://github.com/carbon-design-system/carbon-website/issues/3207#tasklist-block-fafe387e-80d8-4b97-8aa1-9cb6c8766363

PR #3333, opened 2023-01-04
Fluid number input2

aagonzales:
Closes https://github.com/carbon-design-system/carbon-website/issues/3207

Update number input page to include fluid input styles.

Dev: please make sure I did the branching correct.


tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

laurenmrice (changes requested):
Looking great! Just a couple things:

----

### Usage tab

**States section**
- Fixed image:
  - [x] I would bottom align the disabled state container to the others in the row if possible. 
Screen Shot 2023-01-05 at 1 35 58 PM

- Fluid image: 
  - [x] There are cursor state changes here and not in the fixed image. I would align on showing it or not for both images.
  - [x] There is a missing divider rule between the `+/- ` in the Warning state.
  - [x] The labels for the Warning, Error, and Disabled states need to be changed to match the fixed image.
  - [x] Align the numeric value input in the Warning, Error and Disabled states with what you have for the fixed image.
  - [x] Note: These all appear on the style tab Interactive states image tabs too.^
 

### Style tab

**Color section**

- [ ] The image numeric input value of 1000 should show the mono numbers for 0.

**Interactive states section**
- [ ] Need a row for disabled color for controls (-/+).

**Structure section**
- [ ] Fluid input image: Need to use mono 0 for the numeric values in each input.

aagonzales:
@laurenmrice ok I've updated all the images but we decided to use `$body-compact-01` with the number instead of the mono. I doubled checked all the images to make sure they're using the correct token now and updated the type section as well (which I forgot i first round).

PR #3353, opened 2023-01-18
Dropdown accessibility

mbgower:
Update of guidance on Accessibility tab


tay1orjones (approved):

src/pages/components/dropdown/accessibility.mdx
```diff
@@ -6,98 +6,129 @@ description:
...
+<Column colLg={8}>
+
+![Tab to dropdown then operate with arrow keys](images/dropdown-accessibility-1.png)
+
+<Caption>
+
+Dropdowns are in the tab order and operable by Up and Down arrow keys, once
+opened.
+
+</Caption>
```
thyhmdo, line 47: the type format looks correct here but it doesn't look like it's in the caption on the site. Wondering if there should be no space between the Caption and the text.

src/pages/components/dropdown/accessibility.mdx
```diff
@@ -6,98 +6,129 @@ description:
...
-      </StructuredListBody>
-    </StructuredListWrapper>
-  </Column>
+<Column colLg={8}>
+
+![Space and Enter keys open multiselects and toggle the selection of checkbox options. Esc closes.](images/dropdown-accessibility-3.png)
+
+<Caption>
+  Space and Enter keys open multiselects and toggle the state of the checkboxes
+  in the list. Esc closes the list.
```
thyhmdo, line 88: A small thing, but adding "Pressing Esc closes the list" would be consistent with the caption before this.

src/pages/components/dropdown/accessibility.mdx
```diff
@@ -6,98 +6,129 @@ description:
...
+space character into the filter string.
+
+<Row>
+<Column colLg={8}>
+
+![The Enter key opens and selects items in a combo box. Arrow keys can open and navigate the options. Esc clears the input and closes.](images/dropdown-accessibility-4.png)
+
+<Caption>
+  Combo boxes can be opened with the Enter or Up and Down arrow keys. Enter
+  selects the highlighted option in the list. Esc clears the input and closes
```
thyhmdo, line 113: The same thing with adding "Pressing Esc..." would be consistent with the caption before this.

src/pages/components/dropdown/accessibility.mdx
```diff
@@ -6,98 +6,129 @@ description:
...
+opened.
+
+</Caption>
+
+</Column>
+</Row>
+
+#### Dropdown
+
+The default dropdown opens with `Enter`, `Space` or `Down arrow`, with focus
```
laurenmrice, line 54:
```suggestion
The default dropdown opens with `Enter`, `Space`, or `Down arrow`, with focus
```

laurenmrice, line 54:
```suggestion
The default dropdown opens with `Enter`, `Space`, or `Down arrow`, with focus
```

laurenmrice (changes requested):

src/pages/components/dropdown/accessibility.mdx
mbgower: We don't normally use verbs in the captions for these things (since, as you note, we tend to in the surrounding text). It's hardly a fast rule, but given the visual and the preamble, IMO, it's not needed. I do note that one of the captions on the page had the word "press" once. I've removed it for consistency.

src/pages/components/dropdown/accessibility.mdx
mbgower: LOL, okay, so I just made the consistency the other way :) 
I don't think anyone can misread, and whatever can make captions a bit shorter, I tend to be keen to do.

src/pages/components/dropdown/accessibility.mdx
mbgower: I've pushed that change. Just waiting to check the deployment before resolving this conversation.

laurenmrice (approved):

PR #3369, opened 2023-01-26
fix(react): replace codesandbox link with stackblitz

tay1orjones:
Thanks to @skcele for catching this!

#### Changelog

**Changed**

- replaced outdated codesandbox link with the stackblitz vite template link

**Testing / Reviewing**

- Go to `developing/frameworks/react/` on the deploy preview
- Validate the stackblitz resource link works and has the proper icon


francinelucca (changes requested):
Icon is actually not loading for me, not sure why 🧐
image

francinelucca (commented):
I can see the icon now! it looks kind of big to me though 🤔, is this the correct size? 
image

tay1orjones:
@francinelucca I think it's actually the storybook one that is too small - all the icons are within a container of 32x32, but the actual icon size varies. Some don't take up the full space

francinelucca (approved):

aledavila (approved):

PR #3385, opened 2023-02-10
Use Github Script action to sync carbon-platform. branch

francinelucca:
The action we're currently using to sync the `carbon-platform` branch with main doesn't create an intermediary branch which makes it impossible to solve conflicts without modifying main (we don't want to do that).

Plus, the script action is verified :D


tay1orjones (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #3406, opened 2023-02-28
docs(read-only): number input docs

### LINKED ISSUE #3267: [Read-only inputs]: NumberInput website docs

laurenmrice:
Closes #3267 

----

**Changed**

Usage:

- Updated States content
- Updated States images

Style:

- Updated Interactive states table content
- Updated Interactive states images

A11y:

- Updated images to show `body-compact-01` type token


laurenmrice:
- *Note: I revised this image to have 2 states in each row instead of 3. I tried to have an empty block in the very last slot of the image with keeping 3 states per row as we do for normal images, but it sort of breaks the tab image component.
Screen Shot 2023-02-28 at 2 01 06 PM

aagonzales (approved):
It looks good. I'm wondering though if the definition of disabled needs to be cleaned up now so it's very clear and obvious what the difference between the two is (just a thought).

francinelucca (approved):

PR #3440, opened 2023-03-17
(project): update `carbon-website` to node v18

andreancardona:
Closes https://github.com/carbon-design-system/carbon/issues/13322

PR #3456, opened 2023-03-27
(project): update `carbon-website` to node v18

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/13322

At first glance, this PR will look identical to the previous effort @andreancardona and I did in https://github.com/carbon-design-system/carbon-website/pull/3440

The key difference here is that dependencies were updated using node 16 _and then_ updated the project to node 18. This helped avoiding significant issues with transient dependencies not being properly updated to support node 18.

#### Changelog

**Changed**

- update eslint-config-carbon to 3.x and add babel-config-carbon for it (transient dependencies in v2.x did not support node 18)
- update to ignore some new eslint rules that are unnecessary
- update project to node 18


laurenmrice (approved):

andreancardona (approved):
🔥

PR #3457, opened 2023-03-28
docs(about): explain why it is named Carbon

tay1orjones:
[Surfaced in slack](https://ibm.enterprise.slack.com/archives/C0M053VPT/p1680017422882499?thread_ts=1680017422.882499&cid=C0M053VPT)

#### Changelog

**New**

- add a section explaining why the design system is named Carbon

### DIFF
--- a/src/pages/all-about-carbon/what-is-carbon.mdx
+++ b/src/pages/all-about-carbon/what-is-carbon.mdx
@@ -33,6 +33,11 @@ time teams spend designing and building is minimized. Instead of building and
 re-building basic elements, they can spend that time customizing their products
 to address specific client use cases.
 
+The design system is named Carbon, because in nature it builds complex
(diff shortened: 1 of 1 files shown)


tw15egan (approved):

francinelucca:
@aubrey-oneal can we get a content review here? 🙏🏻

aubrey-oneal (dismissed):
It's a little awkward to have 'design system' twice in one clause but I tried rewriting this and couldn't come up with a better option, so I'm good with this. 

But please take out the comma before 'because'. 

"Because is a subordinating conjunction, which means that it connects a subordinate clause to an independent clause; good style dictates that there should be no comma between these two clauses. An exception can and should be made when the lack of a comma would cause ambiguity."

aubrey-oneal (approved):

PR #3523, opened 2023-05-01
[Select] Update doc for select component

### LINKED ISSUE #3458: Select: Usage page
Gower markup: https://ibm.box.com/s/6zkqgvu29ctw52mhfpq7op1l2x413qxj

- [x] Ref issue 1 - https://github.com/carbon-design-system/carbon-website/issues/3361
- [ ] Ref issue 2 - [#13008](https://github.com/carbon-design-system/carbon/issues/13008)
- [ ] Ref issue 3 - [#12721](https://github.com/carbon-design-system/carbon/issues/12721)

> This issue captures the high-level considerations from a review of the Usage, Style, storybook and figma documentation on the Select component. More details can be discovered by examing in the comments in the pdf exports of the [Usage](https://ibm.box.com/s/6zkqgvu29ctw52mhfpq7op1l2x413qxj) and [Style](https://ibm.box.com/s/rs7cqkd5bcdi0dh1d8ptru4acxn1gzdd) pages.

> It is complementary to the https://github.com/carbon-design-system/carbon/issues/13008 issue covering actual implementations with the Select component in the React storybook.

> 1. Various suggested rewrites for clarity and accuracy, commented on in the Usage pdf (linked to above)
> 2. Inconsistencies between images and captions
> 3. Some of the decisions on illustrations between this and the [prior version](https://ibm.box.com/s/5zetn7rglz1tmlzzalhgqy2qcq5lmrxy) of the Usage page are not improvements.
> 4. Calling the select list a "menu" in the anatomy section is ill-advised
> 5. Material in the Content section of the Usage has some challenges.
> 6. "Selected" is missing as a state in the Behaviors section
> 7. There are several inaccuracies in the Style content.

thyhmdo:
Closes #3458 and #3459 

Updated a small amount of content and major issues on images in Usage and Style tabs for Select.


laurenmrice (changes requested):
Looks good Thy! Just a couple comments:

----

### Usage tab

**Select versus Dropdown section:**
- I wonder for this image if the label for the component should say `Select` on the left and `Dropdown` on the right to make it more clear which component is which. Or instead of changing the image, in the caption below it clarify that Select is on the left and Dropdown is on the right.
- For the Dropdown on the right, we do not currently have the prompt in the dropdown field also shown as an option in the dropdown list. Need to clarify with Mike G on this, I think it was another option of how we could structure the Select component, not the Dropdown component.
- For the Dropdown list in this image, the bottom rule under the hover state should not be visible.
Screen Shot 2023-05-04 at 9 28 32 AM

**Validation: Default selection section:**
- I think we should reword this paragraph to be something like below to be more clear on what Carbon offers as a default but what users could modify on their end based on their specific needs:

> The select field has an empty selected option by default, and the user can decide whether to choose a different option from the list. Depending on the use case, this defaulted behavior can be modified to be a prefilled selection to the first option in the list, which is typically shown in alphabetical order, or the prefilled selection could be a frequent or commonly used option that is in the list.

**States section image:**
- Just a small thing, but you can take out the word state in the labels and just have the state name to keep it more consistent and concise with how we are treating our other component state images. This will also apply to the images on the Style tab.
- I think the Disabled state (default style) image should be empty since the rest are empty and it would match the fluid style image.
- The Disabled state label (fluid style) image should say `Disabled`, not read-only.
- The Read-only state (fluid style) should have `$text-secondary` input text.
Screen Shot 2023-05-04 at 9 46 36 AM

----

### Style tab

**Interactive states image:**
- The Disabled state (default style) should not have a border-bottom.
- The Read-only state (fluid style) should have `$text-secondary` input text.

src/pages/components/select/usage.mdx
```diff
@@ -290,49 +290,48 @@ appropriately for its context.
...
 
 #### Helper text
 
 - Helper text is pertinent information that assists the user in choosing the
-  right selection from the select menu.
+  right selection from the select list.
 - Helper text is optional but replaced with warning or error text when these
   states appear.
 
 #### Order
```
laurenmrice, line 306:
```suggestion
#### List order
```
This may be more clear that we are referencing the list option order.

thyhmdo:
@laurenmrice Thanks for the feedback! 

I updated most of the feedback you suggested. There are few things: 
- I liked how you rephrased the part for the paragraphed for modifying the defaulted option. I just took that out from the Validation section since it's down at the bottom and moved it to the Anatomy part instead
- The dropdown example is confirmed by Mike Gower, the only beed he had was the Dropdown selected option is not highlighted enough. He'd prefer better options or ref. [this issue here](https://github.com/carbon-design-system/carbon/discussions/7929#discussioncomment-456319). Though it's not in the code yet so I'm hesitant to update that in the image here. 
- I'd keep the Validation image as it is right now. I think it's not in the State section so it doesn't have to have the state label. It's also about showing an example of validation so maybe it makes sense as it as.

andreancardona (approved):
looks good! 🎉

alisonjoseph (approved):
LGTM!

src/pages/components/select/usage.mdx
```diff
@@ -211,17 +211,21 @@ Inline selects have less visual weight on a page because they are borderless.
 
 1. **Labels:** Text that informs the user what to expect in the list of dropdown
    options.
-2. **Default option:** Usually a frequent option that users would choose from a
-   select list.
+2. **Default option:** An empty option selected by default. The user can decide
```
laurenmrice, line 214: Should the # 2  "Default option" label be called "Selected option" instead? In the image example above, it is just a little confusing because there is already something selected from the list, and it isn't an empty selection which is the default behavior.

src/pages/components/select/usage.mdx
```diff
@@ -290,49 +294,48 @@ appropriately for its context.
...
-should be no longer than three words.
-
 ## Behaviors
 
 ### States
 
 The select has multiple interactive states for both default and inline variants:
-**enabled**, **hover**, **focus**, **open**, **error**, **warning**,
-**disabled**, and **read-only**.
+**enabled**, **hover**, **selected**, **focus**, **open**, **error**,
```
laurenmrice, line 329: There could be a row in the table to represent and describe the `Open state` since it's mentioned in the paragraph above and in the image.

src/pages/components/select/usage.mdx
```diff
@@ -345,7 +348,7 @@ The select has multiple interactive states for both default and inline variants:
 
 <Tab label="Default">
 
-![Select interactive states](images/select-states-fixed.png)
+![Select interactive states](images/select-states-default.png)
```
laurenmrice, line 351: Read-only state (fixed): You can take out the helper text here since you don’t show it for the other states.

src/pages/components/select/style.mdx
```diff
@@ -82,13 +82,13 @@ tabs: ['Usage', 'Style', 'Code', 'Accessibility']
 
 <Tab label="Default">
 
-![Select interactive states](images/select-style-states-fixed.png)
+![Interactive states of default select](images/select-style-interactive-states-default.png)
```
laurenmrice, line 85: Disabled and Read-only state (fixed): You can take out the helper text for these since it is not in the other states.

src/pages/components/select/style.mdx
```diff
@@ -169,7 +169,7 @@ and any proper nouns capitalized. Select text should be three words or less.
 
 <div className="image--fixed">
 
-![Structure and spacing measurements for Inline Select](images/select-style-structure-default-fluid.png)
+![Structure and spacing measurements for fluid select](images/select-style-structure-fluid.png)
```
laurenmrice, line 172: This image is missing the teal padding overlay on the right side for the error state.
Screen Shot 2023-05-10 at 3 24 21 PM

laurenmrice (changes requested):
Just a couple more things!

----

**Overview > Select versus Dropdown section**

- I would mention in this paragraph that you are referring to the `list` styling of the component.
> Another important difference between the two components is the underlying code. The select component’s list appearance will be determined by the browser being used, while the dropdown component’s list appearance can be styled as needed.

- For the dropdown example in this image, use our current dropdown list without the repeated prompt in the list as an option. This isn't implemented yet in code, and we should figure out if that is the direction we want to take with dropdown across the board.

Screen Shot 2023-05-10 at 3 10 41 PM

laurenmrice (approved):

PR #3582, opened 2023-06-06
fix: broken link on other frameworks page

alisonjoseph:


alisonjoseph:
going to include this in another PR with a few more vanilla docs updates

PR #3639, opened 2023-07-11
fix: website button style bugs

### LINKED ISSUE #3637: Color page: tokens tab has overflow menu bug
- On the Color page > Tokens tab, all of the overflows in the table have a dark blue background bug. 

Screenshot 2023-07-11 at 11 42 51 AM

### LINKED ISSUE #3638: UI shell header: App switcher icon has a bug on website
The App switcher icon is not spaced properly within the UI shell header. It currently looks too low. This also applies to the close icon which takes the place of the application icon when the panel is open.

Screenshot 2023-07-11 at 1 55 22 PM

alisonjoseph:
Closes #3638
Closes #3637

#### Changelog

Add style override for switcher / switcher close icon button
Add style to fix overflow menu on color page

### DIFF
--- a/src/styles/_overrides.scss
+++ b/src/styles/_overrides.scss
@@ -7,3 +7,11 @@
 [class^='FourOhFour-module--link']:before {
   color: $text-inverse !important;
 }
+
+// Header icon style fix needed because the Carbon website is on a
+// newer version of Carbon vs. the Gatsby theme
+// If the Gatsby theme is updated to use latest Carbon, this
+// shouldn't be needed
+button[class*='Header-module--switcher-button'] {
+  padding-block-start: 0;
+}
--- a/src/components/ColorTokenTable/color-token-table.scss
+++ b/src/components/ColorTokenTable/color-token-table.scss
@@ -77,6 +77,12 @@
 .color-token-value .cds--overflow-menu {
   margin-left: $spacing-03;
   margin-top: -$spacing-03;
+  padding-block-start: 0;
+  background-color: $layer-01;
+
+  &:hover {
+    background-color: $layer-hover;
+  }
 }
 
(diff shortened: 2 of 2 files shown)


src/styles/_overrides.scss
```diff
@@ -7,3 +7,11 @@
 [class^='FourOhFour-module--link']:before {
   color: $text-inverse !important;
 }
+
+// Header icon style fix needed because the Carbon website is on a
+// newer version of Carbon vs. the Gatsby theme
+// If the Gatsby theme is updated to use latest Carbon, this
+// shouldn't be needed
```
tw15egan, line 14: It looks like `3.1.2` was just released for `gatsby-theme-carbon` which bumps the `@carbon` version to `1.32.0`. Should we try updating that and see if the issue resolves itself?

alisonjoseph: Just checked, both issues are still there 😓

tw15egan (approved):

laurenmrice (changes requested):
- The tooltip should be set to align` top-right `to avoid the off centering and the tooltip needs to be 4px above the overflow menu icon button instead of overlapping with the button.

Screenshot 2023-07-12 at 12 55 09 PM

---

- When the overflow menu is open, the background of the overflow menu icon button should be the same color as the row items. 

Frame 2

alisonjoseph:
@tw15egan The original bugs seem to be gone on the website since the latest update. Thanks for pointing that out @francinelucca  So weird. 

@laurenmrice I'll open a new PR to fix the issues in your comment. I think those are unrelated to whatever else was happening.

laurenmrice:
Oh okay gotcha, thanks @alisonjoseph

PR #3657, opened 2023-07-24
fix: overflow menu and switcher styles

alisonjoseph:
Remove unnecessary @use  statement. 

This should fix the weird blue background on the overflow menu, and switcher styles. 

Double check that `/guidelines/typography/type-sets `page is rendering correctly.


tay1orjones (approved):

tw15egan:
Any idea why this change is causing the grid regressions to appear as well? 

Screenshot 2023-07-25 at 9 09 27 AM

alisonjoseph:
@tw15egan I'm digging into it, seems to be a theme issue so I'll open a PR / release there with latest version of Carbon once I track it down. I rolled the website version back yesterday so the grid issue should be "fixed" for now. (I just updated this branch to reflect that)

tw15egan (approved):

PR #3724, opened 2023-08-30
docs: fix a11y tab do image size

laurenmrice:
Fixes the size of the "do" images on the A11y tab for the Tabs component. They should be 6 columns wide.

----

**Changed**

- Column size for "do" images


laurenmrice:
@alisonjoseph or @tw15egan Would either of you be able to help me with this?

I am trying to do 6 column "do" images, like we do in the [Tooltip component](https://carbondesignsystem.com/components/tooltip/usage/#when-not-to-use), however, the images in the deploy preview do not update. 

I also noticed that this PR has merge conflicts with the images I am revising. I tried to rename them/delete them and re-add them to the PR, but I can't get the conflicts to go away.

alisonjoseph:
@laurenmrice @tw15egan  I can take a look

alisonjoseph (approved):
Should be good now I think

aagonzales (approved):
Looks good!

Issue #3519, opened 2023-05-01
Component Accessibility Status: Final Dev Deliverable


laurenmrice:
**Developer** | Will work on showing this final design on the website while working closely with a designer (Thy) 

- [ ] Task 1 (dev defines the work here)
- [ ] Task 2 (dev defines the work here)
- [ ] Task 3 (dev defines the work here)

tay1orjones:
As part of this, dev should explore how to surface reporting of what types of tests a component has via playwright's `grep` tags. We may need to add new tags, `@avt`, `@avt-complex-states`, `@avt-keyboard-nav`, `@vrt`, etc.

By gathering these tests per-component we should be able to produce a JSON file or something that contains data giving us the ability to build a table something like

| component | `@avt` | `@avt-complex-states` |
|--------|--------|--------|
| Dropdown | ✅ | 🚫 |
| MultiSelect | ✅ | ✅ |
| Button | 🚫 | ✅ |

tay1orjones:
Another concern brought up was that as we look at this, we should take into account that these tests are only for `@carbon/react` and the design needs to scale to include other framework implementations, such as `@carbon/web-components`.

tay1orjones:
Looking forward, this issue will contain the work that comes out of the design explorations:

* https://github.com/carbon-design-system/carbon/issues/14484
* https://github.com/carbon-design-system/carbon-website/issues/3518

sstrubberg:
closing in favor of the issues created yesterday in refinement.

PR #3735, opened 2023-09-06
feat(avt): surface component avt test status

tay1orjones:
Part of #3519 
Also part of https://github.com/carbon-design-system/carbon/issues/14480

This PR updates the component overview page to display a new table that surfaces the avt testing status of every component. 

This is just a walking skeleton and a demo of the type of things we _could_ do. It only works for Accordion right now. The rest of the components have avt tests, but mistakenly display 🚫 for now because their avt tests have not yet been updated to use the specific naming convention for the tags that we've established.

Overall this will probably be just the beginning of the design explorations of how we surface these on the website.

---

### Outstanding items that I'd like to iterate further

- [ ] Include a link to the test file source somewhere in the table. The data coming from `@carbon/react` includes the name of the test file and the line number for each individual test. Users could observe the table and be linked directly to the test corresponding to the "✅"
- [ ] Update the remaining tests in the monorepo to use the tag naming convention we've established


Huixh111:
o

PR #3759, opened 2023-09-28
fix: tooltip docs typo and bug

laurenmrice:
**Changed**

- Fixed a sentence where there was incorrect information about screenreaders.
- Fixed a spacing bug with a heading being too close to the live demo, causing it to stylistically break.


alisonjoseph (approved):

kingtraceyj (approved):
good to go! 💥

PR #3782, opened 2023-10-18
docs: update releases page

janhassel:
I found the current releases page to be a bit confusing for the following reasons:
- The preferred `@carbon/react` package isn't listed
- The `carbon-components-react` package is called `Carbon React` which sounds like it's referring to `@carbon/react`
- Some of the links were pointing to the wrong versions

#### Changelog

**Changed**

- On "All about Carbon -> Releases -> Release History"
  - Renamed column headers to use actual package names
  - Added column `@carbon/react`
  - Updated version links if possible[^1]

[^1]: Is it intentional, that `carbon-components@latest` directs to `10.x` and `carbon-components-react@latest` points to `8.x`? I'd imagine these versions are incompatible and could mess with one's initial setup.
Maybe this could be updated and the following tags could be created in which case they could be linked in this page: `@carbon/elements@10.x`, `carbon-components@10.x`, `carbon-components-react@7.x`


laurenmrice (approved):

alisonjoseph (approved):
LGTM!

PR #3944, opened 2024-02-26
docs: Vertical dividers doc update (Search fluid)

alina-jacob:
Closes [#15753](https://github.com/carbon-design-system/carbon/issues/15753)
[Figma Ref File](https://www.figma.com/file/oURkNA2HJwIWKIzLj8Q159/Alina's-Workspace?type=design&node-id=1097-50034&mode=design)


aagonzales (commented):
@alina-jacob The images look good but you've exported them too large. The original size is 608px wide and we 2x export those to have a final size of 1216px wide. It looks like you might have started with 1216px and then 2xed that.  

Proportionally they look correct and wouldn't be a big deal except for the file size memory that larger images take up.

aagonzales:
@alina-jacob oh you did 4x that also explains it. Yes, they just need to be exported at 2x.

aagonzales (approved):

laurenmrice:
@alina-jacob I left a comment in your Figma file about the two do images. They currently blend in with the Gray 10 background of the website page because the image also have Gray 10 backgrounds, we just need to add the borders to the images. Other than that looks great!

guidari (approved):

Issue #3914, opened 2024-02-13
Interactive tag: Accessibility tab docs


laurenmrice:
### Acceptance criteria
- Update the Accessibility tab to use th [accessibility content template](https://carbondesignsystem.com/contributing/documentation/#accessibility-template).
- Add specific content and images for the new additions of the Interactive tags (Dismissible, Selectable and Operational)

PR #3946, opened 2024-02-27
docs: update tag docs on a11y tab

laurenmrice:
Closes [#3914](https://github.com/carbon-design-system/carbon-website/issues/3914)

Updates the Accessibility tab docs on the website for the Tag component with the content template and new interactive tag variants. 

#### Changelog

**New**

- Added images and content for interactive tags. 

**Changed**

-  Changed the content structure of the page.


guidari (dismissed):
LGTM! 🚀

guidari:
Also other the thing we came across while developing the interactive tag is the problem of nesting buttons. 
For example when we start to code the `Oprational Tag` using a `ToggleTip`, both are buttons and shouldn't be nested.

```
  <ToggletipButton label="Tag content" tabIndex={-1} as="div">
    <OperationalTag
      renderIcon={Asleep}
      className="some-class"
      {...args}>
      {'Tag content'}
    </OperationalTag>
  </ToggletipButton>
```

So we had to use `PolymorphicProps` to change the element tag. If a user wants to create a custom component it shouldn't nested buttons.

laurenmrice:
Just some notes about these comments:

Yes, all variants of Tag have their own distinct visual differences before interacting with them. 

description

---

This tag style will be updated to use the correct styling of the tag we have in code. We did not have that new variant in code when the A11y table was created. 

tag

---

I took this part out. That part was trying to say that operational tags can be triggered to open a view of tags, and it can do this through different ways by either using a popover, a toggletip, a modal, etc. I didn't want to get into the details of all of the different types of interactions and keys that could be used in these scenarios because it could be a lot of information?

Screenshot 2024-04-08 at 8 44 01 AM

mbgower (approved):
This looks good. We are having a discussion on methods for exposing truncated tag strings, but I think what is captured here makes sense.

I took a moment to review the Usage tab as well (although I realize it is not in scope for this PR). 
It seemed to support the material written in the a11y tab.

The one thing I found a little odd was having a disabled state for the read-only tag, since by definition, a read-only tag is no operable. 
>Read-only tags: enabled, disabled, and skeleton

I'm assuming that if the tag was on a component such as a tile that was itself disabled, the idea that a read-only tag on a disabled tile would likewise have its text greyed out.

laurenmrice:
I am taking out the overflow section for now, until we can get that finalized. I will make a separate PR in the future to add overflow guidance to the different tabs for the component once it is finished in code.

src/pages/components/tag/accessibility.mdx
```diff
@@ -19,205 +19,70 @@ import {
...
-      </StructuredListBody>
-    </StructuredListWrapper>
-  </Column>
-</Row>
+Keep this in mind if you’re modifying Carbon or creating a custom component.
+
+- Do not add an `onClick` functionality to the dismissible tag, and only reserve
+  interactions for the close icon in the tag.
+- Do not nest buttons within the operational tag. Consider using
+  `PolymorphicProps` to change an element tag to avoid nesting buttons.
```
guidari, line 88: I would write like that: 
```suggestion
- Do not nest buttons within the operational tag. Consider using
  `PolymorphicProps` to change an element tag to avoid nesting buttons.
```

guidari (approved):
LGTM!

src/pages/components/tag/accessibility.mdx
laurenmrice: Thank you!

src/pages/components/tag/accessibility.mdx
```diff
@@ -19,205 +19,70 @@ import {
...
-    (WCAG Success Criteria
-    [1.4.1](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color))
-  - [1.4.3 Contrast (Minimum)](https://www.ibm.com/able/requirements/requirements/#1_4_3)
-    (WCAG Success Criteria
-    [1.4.3](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum))
+</Column>
+</Row>
 
-## Accessibility testing
+Focus is shown around each selectable tag. Pressing `Enter` or `Space` toggles
```
thyhmdo, line 59: I think this is missing a short intro to the selectable tag. For example: 
- Read-only...
- Dismissable tag...
- Selectable tag...
- Operational tag...

Could be as repetitive as the Dismissable tag, but I think it's helpful to know the tab order receiving focus on the entire tag button.

src/pages/components/tag/accessibility.mdx
```diff
@@ -19,205 +19,70 @@ import {
...
-          </StructuredListCell>
-        </StructuredListRow>
-      </StructuredListBody>
-    </StructuredListWrapper>
-  </Column>
+</Column>
 </Row>
 
-### macOS Screen reader tests
+Focus is shown around each operational tag. Pressing `Enter` or `Space` will
```
thyhmdo, line 70: the same comment here for a short intro sentence.

src/pages/components/tag/accessibility.mdx
laurenmrice: I did have that before but removed it. I can add it back

thyhmdo (approved):
Looks great!

PR #3995, opened 2024-04-02
Fixes the Tag LiveDemo

sstrubberg:
This is busted.

https://carbondesignsystem.com/components/tag/usage/#live-demo

This should show a working LiveDemo for `Tag`

https://carbondesignsystem-git-sstrubberg-patch-1-carbon-design-system.vercel.app/components/tag/usage/#live-demo


Kritvi-bhatia17 (approved):
Good catch @sstrubberg 🔍

PR #4184, opened 2024-07-30
Update accessibility.mdx

### LINKED ISSUE #3960: The Accessibility testing status section of the Accessibility tab should follow the Development considerations
## Summary
The new information for accessibility testing is very welcome, but it could be better positioned and potentially needs its heading hierarchy addressed

## Relevant information
The new accessibility testing information appears at the top of the Accessibility tab for each component.

While this information will be valuable to some, it is arguably less valuable to designers (the primary target group for the Accessibility tab) than the previously existing information on the page (What Carbon provides, Design recommendations, Development considerations). This is especially the case since the accessibility summary has already been surfaced near the top of the Usage page. As well, activating any of that summary information on the Usage page relocates the user to this section. In other words, the content is already highly discoverable without being placed at the start of the accessibility information.

A concern with it preceding the main guidance intended for designers ("What Carbon provides") is that designers are less likely to identify the key concepts on the page, when confronted with testing information that may not be pertinent to their role.

I recommend it be placed at the bottom of the page. Its heading could be listed as part of the in-page TOC links at the top (from which it is currently absent), like so:
[What Carbon provides](https://carbondesignsystem.com/components/form/accessibility#what-carbon-provides)
[Design recommendations](https://carbondesignsystem.com/components/form/accessibility#design-recommendations)
[Development considerations](https://carbondesignsystem.com/components/form/accessibility#development-considerations)
[Accessibility testing status](https://carbondesignsystem.com/components/form/accessibility#accessibility-testing-statusobject-objectobject-object)

Finally, it is currently presented as an H3. Especially given its current position, this is a hierarchical failure, since it precedes the first H2 on the page. At the least, if left in this position, it needs to be promoted to an H2. If the testing information is relocated to the bottom of the page, one can make the case that the testing status is a child of the Development considerations, but I think it can also work as a peer.

# Conclusion from refinement 
We agree and we will move the testing section to the bottom ~~and add an H2 heading~~ [decided on Aug 6 call to leave as H3, mg]. Do this for all component pages. ~~The heading may be a part of the component so this may need to be a dev issue. ~~ [Since leaving as H3, unnecessary]

mbgower:
Tried moving the accessibility status to the bottom of the page, as per #3960 

Since the H3 is generated in code, I'm just leaving it as is. This becomes a subsection of Development considerations, which works fine. 
https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/accordion/accessibility

Closes #3960


alisonjoseph:
If we're moving this section to the bottom (and updating to a h2) we probably should add an anchor link for "Accessibility testing status" to the top of these pages also.

thyhmdo (approved):
Looks good to me.

Kritvi-bhatia17 (commented):
Hi @mbgower! This looks good 🎉 
Just a quick confirmation—will the change in this PR need to be updated in all the other components as well?

alisonjoseph (approved):

laurenmrice:
@Kritvi-bhatia17 @mbgower Yes, I believe he said he would make the change in this PR for all of them, once we agreed on a direction. So we should wait to merge.

mbgower:
@laurenmrice I've now repositioned on all relevant pages, and cleaned up a couple of in-page links at the same time.
The only other thing we _may_ want to consider doing with this PR is that there are several pages that were never updated to the new accessibility tab template (because the component was scheduled to be updated/replaced). These have an Accessibility testing H2. Repositioning still works on these pages, since the new material complements the legacy testing info on these pages; however, if you want I could strip out the legacy test stuff. Those pages are:

- [date picker](https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/date-picker/accessibility#accessibility-testing)
- [notification](https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/notification/accessibility#accessibility-testing)
- [progress indicator](https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/progress-indicator/accessibility#accessibility-testing)
- [structured list](https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/structured-list/accessibility#accessibility-testing)
- [tile](https://carbondesignsystem-git-mbgower-patch-1-carbon-design-system.vercel.app/components/tile/accessibility#accessibility-testing)

laurenmrice:
@mbgower This looks great. Thank you! If you don't mind doing it in this PR, it would be good to strip those components of the legacy content for now. I also noticed that [List](https://carbondesignsystem.com/components/list/accessibility) does not have a page header.

laurenmrice (approved):
Looks good! Thank you!

Issue #4180, opened 2024-07-25
[Content templates] Style template: for components with multiple variants


aagonzales:
## Acceptance criteria
Add a new [Style template](https://carbondesignsystem.com/contributing/documentation/#style-template), for components with multiple variants, with the following structure:

### Multiple variants 

- **Intro**
- **Anchor links**
- **Color**
   - Variant 1 
      - Style A / Style B (optional)
      - Interaction states
   - Variant 2 
      - Style A / Style B (optional)
      - Interaction states
- **Typography**
- **Structure**
   - Variant 1
      - Style A / Style B (optional) 
   - Variant 2
      - Style A / Style B (optional) 
- **Sizes** (optional) 
   - Variant 1
      - Style A / Style B (optional) 
   - Variant 2
      - Style A / Style B (optional) 
- **AI Presence** (optional)
   - Variant 1
      - Style A / Style B (optional) 
   - Variant 2
      - Style A / Style B (optional)

Kritvi-bhatia17:
Hi @aagonzales! This is looking good 🔥

Just confirming a few minor things:
1. Is it "Interactive states" or "interactive states"?
2. Should it be "Sizes" or "Size"?
3. Will we be including a 'Feedback' section on the Style tab?

PR #4197, opened 2024-08-13
(docs) Content template: updated style sections

aagonzales:
Closes https://github.com/carbon-design-system/carbon-website/issues/4180

Updated the content template to include a section for `Style tab with multiple variants`

#### Changelog

**New**

- Added "Style template: for components with multiple variants" 
- Added an "AI presence" template section to both "one variant" and "multiple variants" sections (style tab only)

**Changed**

- Updated "Style template: for components with one variant" to be exclusive about a single variant. I removed any information that was talking about variant structure. 
- Reformatted the instructional comments to be more skimmable.


laurenmrice (approved):
I think this looks really good, and its working well for my Pagination Style tab docs! 

Going to approve and I think one thing you could add to the Typography section is to be more explicit about how to treat different variants that have different type specs, since Pagination ran into that issue.

Kritvi-bhatia17 (commented):
Hi @aagonzales! This looks great. 🔥
I just wanted to confirm one small thing: the current docs mention "interactive state," but in the updated document templated, it says "interactive state color." Which one is correct? As this could lead to changes in almost all the components' docs.

Screenshot 2024-08-16 at 12 11 04 PM
Screenshot 2024-08-16 at 12 18 00 PM

aagonzales:
@Kritvi-bhatia17 oh yes good catch. I think Aubrey changed it to "interactive state color" the last time she updated the template so I just kept it that way. You are right to point out its inconsistent with what's showing in most docs. We probably need to do a content scrub to make sure all our style tabs will reflect the new template.

Kritvi-bhatia17 (approved):
LGTM! 🎉

alisonjoseph (approved):

PR #4200, opened 2024-08-15
fix: update ai label callout to explainability popover in demo

alisonjoseph:
#### Changelog

**Changed**

- update callout>explainability popover in live demo for AiLabel


aagonzales (approved):

emyarod (approved):
looks good to me!

PR #4260, opened 2024-09-12
chore(husky): fix pre-commit hook

kennylam:
Update `husky` pre-commit hook to prevent it trying to run commands that don't exist.

#### Changelog

**New**

- add `prepare` script for Husky

**Changed**

- update husky command in `pre-commit` script

**Removed**

- `husky` entry in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "format": "prettier --write 'src/**/*.{css,scss,json,html,yaml,md,mdx,js}'",
     "format:check": "prettier --check 'src/**/*.{css,scss,json,html,yaml,md,mdx,js}'",
     "update-browserslist": "npx browserslist-ga",
-    "postinstall": "husky"
+    "prepare": "husky"
   },
   "engines": {
     "node": ">=18.x"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -19,7 +19,8 @@
     "lint:js": "NODE_ENV=test eslint . --fix",
     "format": "prettier --write 'src/**/*.{css,scss,json,html,yaml,md,mdx,js}'",
     "format:check": "prettier --check 'src/**/*.{css,scss,json,html,yaml,md,mdx,js}'",
-    "update-browserslist": "npx browserslist-ga"
+    "update-browserslist": "npx browserslist-ga",
+    "postinstall": "husky"
```
tay1orjones, line 23: The [latest release](https://github.com/typicode/husky/releases/tag/v9.0.1) suggests this should be on `prepare` instead of `postinstall`, what do you think?

kennylam: Ah ok cool. I actually had it `prepare` at first but wanted to be consistent with what was being used elsewhere. But I'll change it 👍

annawen1 (approved):
Works for me locally - was able to commit with no issues!

aagonzales (approved):

PR #4281, opened 2024-09-23
chore: delete tsconfig.json

alisonjoseph:
File is no longer needed as we don't have any ts files.


Kritvi-bhatia17 (approved):
Thanks for promptly fixing this, Alison! LGTM 🚀

Issue #3935, opened 2024-02-21
 [Dropdown: Combobox & Multiselect]: Update website docs with focus state changes


Kritvi-bhatia17:
Related [#768](https://github.com/carbon-design-system/carbon-design-kit/issues/768#issue-2146053598)

## Acceptance criteria
Update the Dropdown focus states on website doc with what came out of our [final design solutions](https://github.com/carbon-design-system/carbon-design-kit/issues/768#issue-2146053598).

Ref: [Figma link](https://www.figma.com/file/YAnB1jKx0yCUL29j6uSLpg/(v11)-All-themes---Carbon-Design-System?type=design&node-id=14032-290630&mode=design)

---

### Style tab

- [x] Update all the Images under the Stye tab

---

### Accessibility tab

- [x] Update all the Images under the Accessibility tab

---

### Review

Image review
- [x] Carbon designer (Thy/Lauren)
- [x] Accessibility review (Michael)

PR 
- [x] Create PR
- [ ] PR review

thyhmdo:
Kritvi needs to revaluate this issue

thyhmdo:
Waiting on Benjamin to raise the PRs

sstrubberg:
Update these screens with what needs to be actually changed. @Kritvi-bhatia17 

Remove / replace with only the needed screens.

Kritvi-bhatia17:
Increasing the story points for this issue from 5 to 8, as it required creating and updating all images under both the Style and Accessibility tabs following the latest focus state discussion.

PR #4342, opened 2024-10-29
fix: Updated images in the Dropdown's Style and Accessibility tabs to display the focus state

### LINKED ISSUE #4049: [Dropdown: Combobox] Separate image needed for Combobox under Interactive states
## Acceptance criteria

In the style tab, under interactive states, there's currently a single image illustrating both the dropdown and combobox components merged, linked [here](https://arc.net/l/quote/rqvclrsn).

Given that dropdowns and comboboxes now have distinct states and behaviors, we require a separate image specifically for the combobox.

### Tasks

- [x]  Create an image for the Combobox interactive section.

Review
- [x] Carbon designer (Thy)

PR 
- [x] Create PR
- [ ] PR review

---

### Additional notes

[Image link](https://carbondesignsystem.com/components/dropdown/style/#interactive-states)

image

Kritvi-bhatia17:
Closes #4049 & #3935  

This PR reflected the updated images with the focus stroke change and add a new image for Combobox on the Style tab.

#### Changelog

**New**

- Created an image for the Combobox interactive section.

**Changed**

- Updated the images under the Style & Accessibility tabs to reflect the latest focus state discussion.


laurenmrice (changes requested):
## Style tab

### Color section
- Shouldn’t the open states have the focus on the fields?
- Combo box should be two words.
- The caption is wrong here. It should be something like “Color states of dropdown, combo box, and multiselect.”

### Interactive states section

#### Dropdown images
- The default image has the wrong label color for the “Warning” and “Error” states. It should be `$text-secondary`.
- For the default and fluid images, shouldn’t the “Hover open” state have the focus state on the field?

#### Combo box images
- In the default and fluid images for the “Selected” states and the "Read-only" state, the third option should be selected for consistency.

#### Multiselect images
- The “Disabled” and “Read-only” state is the wrong size. It should be Medium like the others.
- For the default and fluid images, shouldn’t the “Hover open” state have the focus state on the field?
- For the default and fluid images “Hover open” state, the hovered option text should be `$text-primary`.
- For the default image, the “Read-only” state should have "3" in the tag for consistency.
- For the fluid image, the “Warning” state message text should be `$text-primary`.
- For the default and fluid images in the “Selected open” states, I am pretty sure the word “All” is supposed to be `$text-secondary` even if all options are selected in the menu list.
- For the default and fluid images in the "Focus open" state, the dividers should be inset for the options in the menu. It is only the divider for the All option that is flush.
- I couldn't make an inline edit for you for this one, but for the caption below the Multiselect interactive state images, the caption should say "Multiselect interactive states".

**Inline images**
- Shouldn’t the “Hover open” state have the focus state on the field?
- The open menus are not all the same width. 
- We might need to restructure this image a bit to balance it out, I can take a look at it.

laurenmrice (changes requested):
## Accessibility tab

### Keyboard interactions section
 
#### Section image
- Shouldn’t the focus state be on the option in the menu since it is selected?

#### Multiselect image 
- The chevron needs to be pointing "up" since the menu is open.

src/pages/components/dropdown/style.mdx
```diff
@@ -140,7 +140,7 @@ structure, and size and AI presence.
 </Column>
 </Row>
 
-<Caption>Dropdown and combo box states</Caption>
+<Caption>Dropdown states</Caption>
```
laurenmrice, line 143:
```suggestion
<Caption>Dropdown interactive states</Caption>
```

Just to be more clear that these are interactive states.

src/pages/components/dropdown/style.mdx
```diff
@@ -149,13 +149,37 @@ structure, and size and AI presence.
 
 <Tab label="Default">
 
-![Default multiselect dropdown states](images/dropdown-multiselect-style-default-states.png)
+![Default combo box dropdown states](images/combobox-style-default-states.png)
```
laurenmrice, line 152:
```suggestion
![Default combo box interactive states](images/combobox-style-default-states.png)
```

We typically don't refer to it as "combo box dropdown" and instead we just say "combo box" to not confuse people with the "dropdown" variant.

src/pages/components/dropdown/style.mdx
```diff
@@ -149,13 +149,37 @@ structure, and size and AI presence.
...
 
 </Tab>
 
 <Tab label="Fluid">
 
-![Fluid multiselect dropdown states](images/dropdown-multiselect-style-fluid-states.png)
+![Fluid combo box dropdown states](images/combobox-style-fluid-states.png)
```
laurenmrice, line 158:
```suggestion
![Fluid combo box interactive states](images/combobox-style-fluid-states.png)
```

src/pages/components/dropdown/style.mdx
```diff
@@ -149,13 +149,37 @@ structure, and size and AI presence.
...
+
+</Tab>
+
+</Tabs>
+
+</Column>
+</Row>
+
+<Caption>Combo box dropdown states</Caption>
```
laurenmrice, line 167:
```suggestion
<Caption>Combo box interactive states</Caption>
```

src/pages/components/dropdown/style.mdx
```diff
@@ -149,13 +149,37 @@ structure, and size and AI presence.
...
+
+<Row>
+<Column colLg={12}>
+
+<Tabs>
+
+<Tab label="Default">
+
+![Default multiselect dropdown states](images/multiselect-style-default-states.png)
```
laurenmrice, line 176:
```suggestion
![Default multiselect interactive states](images/multiselect-style-default-states.png)
```

src/pages/components/dropdown/style.mdx
```diff
@@ -149,13 +149,37 @@ structure, and size and AI presence.
...
+
+</Tab>
+
+<Tab label="Fluid">
+
+![Fluid multiselect dropdown states](images/multiselect-style-fluid-states.png)
```
laurenmrice, line 182:
```suggestion
![Fluid multiselect interactive states](images/multiselect-style-fluid-states.png)
```

laurenmrice (changes requested):

src/pages/components/dropdown/images/combobox-style-default-states.png
mbgower, line 1: When opened, the selected option should have a tick beside it

mbgower:
> ## Style tab
> ### Color section
> * Shouldn’t the open states have the focus on the fields?

@laurenmrice this is really a Carbon decision, not an accessibility MUST do. What I believe Kritvi has tried to make consistent is that focus is retained in the parent in two aligned circumstances:
1. There is no option selected
2. Either a _mouse_  or the Enter key is used to open the component.

If _either_ a DownArrow is used to open the component _or_ there is an option selected, focus moves to the option list.

I emphasize that this ultimately is a design decision. You could as easily say that focus _always_ moves to an item in the options when the component is opened, regardless of method. I believe that is more historically how Carbon did things. It is the pattern ARIA has for their [Listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) pattern. And it is very similar to the [Menubutton](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) interaction, which is a similar _but different_ disclosure pattern, with a much more constrained interaction model.

However, Kritvi's approach, based on a common implementation in Storybook, is to consistently apply the above numbered behaviour. I believe it is defendable as an approach for any kind of non-menu 'dropdown' disclosure: for a mouse user, the focus remains on the trigger; for a keyboard user, Enter is used as an open/close toggle while DownArrow is used as a open/move focus to child operation. This aligns, to a degree, with the APG [Disclosure](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) pattern.

laurenmrice:
@mbgower My question was that the image here seems wrong because there is no focus state when the dropdown is open, and there is no selection. My understanding is that there needs to be focus somewhere, whether it be on the field/trigger, or on the first option in the menu. My understanding is that for mouse users, design has decided that if the menu is open and nothing is selected, the focus would appear on the field/trigger.

Screenshot 2024-10-30 at 2 04 54 PM

Kritvi-bhatia17:
@laurenmrice I'll update all images on the Style tab to reflect the latest focus state discussion.

laurenmrice:
@Kritvi-bhatia17 let me know when this is ready for review again

Kritvi-bhatia17:
Hi @laurenmrice, this is ready for review, and I’ve already requested a re-review from both you and @mbgower. Thank you!

> @Kritvi-bhatia17 let me know when this is ready for review again

Kritvi-bhatia17:
@laurenmrice I've already made the changes, and they’re reflected in one of the commits, though I'm not sure why it is not showing up. Either way, I’ve addressed all the comments again.

laurenmrice (changes requested):
@Kritvi-bhatia17 I think after these changes, it will be good to go! 

---

### Color section
- For the default image, the Multiselect closed components is the large size, it needs to be the medium size to match all of the other component sizes in this image for consistency. 
Screenshot 2024-11-07 at 8 08 21 AM


### Interactive states section

#### Multiselect
- Something I did not catch before, but the default and fluid images error state error border should be 2px not 1px.

#### Inline
- Something I did not catch before, but the read-only state input text “Option 3” should be text color `$text-primary`.

laurenmrice (approved):

alisonjoseph (approved):
LGTM!

Kritvi-bhatia17:
Hi @mbgower, could you please review this PR? Thanks!

mbgower:
> Hi @mbgower, could you please review this PR? Thanks!

Rather than look at just the changes in the PR, I looked at the whole a11y tab in the preview. Within that context, I have some concerns not specific to this PR; the keyboard guidance seems unchanged. For example, it still reads:

>Dropdown
The default dropdown opens with Enter, Space, or Down arrow, with focus moving to the currently selected option. Arrow keys change options, and any option with focus is selected with Space or Enter, which also closes the dropdown. Pressing Esc or Tab closes a dropdown without changing the selected option.

I believe this isn't precise enough, given the interaction model you have championed, which I [describe in a prior comment](https://github.com/carbon-design-system/carbon-website/pull/4342#issuecomment-2447854966).

To me, either the accessibility tab or the usage tab must must explain the interaction model for focus handling. The above quoted paragraph is one place it could potentially be included, although you then need to address it somehow for all the variatns. Another option is to have a separate subsection called something like 'Focus handling in open dropdowns'.

I am generating a PR to address.

mbgower:
See https://github.com/carbon-design-system/carbon-website/pull/4362

PR #4374, opened 2024-11-21
[Inline loading] Updated text size and states

thyhmdo:
Closes #https://github.com/orgs/carbon-design-system/projects/39/views/13?filterQuery=sprint%3A%22%40next%22+design&pane=issue&itemId=80437243&issue=carbon-design-system%7Ccarbon-website%7C4271

**New**

- Updated the text size to 14pt 
- Updated the inactive state (to match with the code)


laurenmrice (changes requested):
- This image needs the borders on the edges so the background doesn't blend in with the website page background.

Screenshot 2024-11-25 at 12 33 24 PM

----

- Could you try to vertically align the labels of these and visually center them within each section? Would be good to update the Style image to this too.
- The divider borders look like they are two different colors. They should be the border-subtle-01 color for the gray 10 theme. 

Screenshot 2024-11-25 at 12 34 24 PM

----
- The inline loading label should be $body-compact-01.
Screenshot 2024-11-25 at 12 36 28 PM

guidari (approved):

laurenmrice (changes requested):
Hey @thyhmdo ! The dividers here need to be the $border-subtle-01 in the Gray 10 theme.  Right now they are using the White theme and are a bit dark. It should be changed on both the Usage and Style tab. Other than that it looks good!

Screenshot 2024-12-04 at 6 07 39 PM

laurenmrice (approved):

PR #4394, opened 2024-12-10
fix(issue-triage): add pat

ariellalgilmore:
missing personal access token config

### DIFF
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -41,13 +41,11 @@ jobs:
             echo "Issue ID is $ISSUE_ID"
             echo "issue_id=$ISSUE_ID" >> "$GITHUB_OUTPUT"
         env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          GH_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
       - name: Update issue with website area field
         run:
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-triage.yml
```diff
@@ -42,10 +42,12 @@ jobs:
...
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
       - name: Update issue with website area field
         run:
           gh project item-edit --project-id $PROJECT_ID --id $ISSUE_ID
           --field-id $AREA_FIELD_ID --single-select-option-id $AREA_WEBSITE
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
tay1orjones, line 52: Should we just rip out the app token and exclusively use the PAT token instead?

```suggestion
          GH_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```

I'm not sure the scopes that one has, but we can try it. We also have `ADD_TO_PROJECT_PAT` in the secrets that for sure has the projectsV2 scope but idk if it has the others necessary for the repo/issue queries.

ariellalgilmore: updated to use the `ADD_TO_PROJECT_PAT`

tay1orjones:
So wild that they still haven't fixed this wrt needing a PAT over an app token: https://github.com/orgs/community/discussions/46681#discussioncomment-8774842

tay1orjones (approved):

Issue #4375, opened 2024-11-21
Update of developer documentation on the Carbon website


sstrubberg:
Some thoughts before we start converting these to tasks:

- We should segregate the community initiatives from the officially supported ones. E.g. Angular, Vue, etc shouldn't be on the same level as React and Web Components
- Web Components could use a primer to describe what they are, so folks that are both technical and non-technical can understand what they are.
- Some specific topics around React itself could be added. @tay1orjones mentioned things like server-side components, and our stance on that.
- With these items, something to consider is if there could be a better IA to support this content, perhaps can consider taking advantage of treeview navigation if it makes sense.
- Other topics we can potentially surface: feature flags, LTS, CVE mitigation/security support, etc

sstrubberg:
https://github.com/carbon-design-system/carbon/issues/16985

PR #4418, opened 2025-01-02
docs(developer): official vs community content

jeffchew:
Closes #

https://github.com/carbon-design-system/carbon-website/issues/4375

This is a first pass at differentiating between officially supported frameworks and community content. This is a good starting point, and we can iterate on this and think about better IA especially if we decide to bring in things like tree view navigation.

#### Changelog

**Changed**

- Multiple changes in developer documentation


src/data/nav-items.yaml
```diff
@@ -27,16 +27,12 @@
   pages:
     - title: Get started
       path: /developing/get-started/
-    - title: Frameworks
+    - title: Frameworks (official)
       path: /developing/frameworks/react/
+    - title: Frameworks (community)
```
tay1orjones, line 32: Small suggestion

```suggestion
    - title: Frameworks
      path: /developing/frameworks/react/
    - title: Community frameworks
```

jeffchew: Thanks for the suggestion! I made the changes across the rest of the subpages too.

PR #4433, opened 2025-01-16
fix: fixed live demo

guidari:
Fixed live demo path

Working after the path fix

Screenshot 2025-01-16 at 17 38 46


alisonjoseph (approved):

emyarod (approved):
looks good to me!

PR #4446, opened 2025-01-28
doc(toggle): Update the Style tab with the latest content template

### LINKED ISSUE #4216: [Toggle]: Style tab
## Acceptance criteria
Update the Style tab with the style [content template](https://carbondesignsystem.com/contributing/documentation/#style-template:-for-components-with-multiple-variants).

----

- [x] Convert all the "action text" to "state text", both in images and texts (Michael's explanation below)
- [x] Change "Label" to "Reminder" in all the images where needed.
- [x] In general for the tables on this page, "text color" needs to have a dash instead of a space. It should read "text-color".

### Interactive states
*Just noting this current information on the website needs to be updated, then you can move it to the proper sections according to the new template and remove this section.
- [x] There are two color tokens in this table that do not have the $ before the token name, but they need it.

- [x] The description under the table needs to have a * before the word "Denotes" instead of a dash.
  > Fixed in this PR #4435. 
- [x] The image caption should not have a period after it.

# Color

> ### Default toggle color
> - Add this section
> - Add a table
> - Add an image
> 
> #### Default toggle interactive state color
> - Add this section
> - Add a table
> - Add an image
> 
> ### Small toggle color
> - Add this section
> - Add a table
> - Add an image
> 
> #### Small toggle interactive state color
> - Add this section
> - Add a table
> - Add an image

> [!NOTE] 
> **My observation:** 
Both variants share the same table for colors and interactive states, meaning the exact same table would be repeated. It might be better to combine them, but would need to see how it looks laid out before deciding—this is something I’ll present during the design critique call.
Additionally, we should consider how this approach impacts other sections with variant-specific content. It might be worth revisiting the documentation guidelines to account for exceptions like this. For instance, in the Tiles Usage documentation, the guidelines were adjusted to include anatomy under each variant rather than under the general 'Formatting' section.

- [x] Playback to the team 
> Notes from the Playback:
It’s fine to combine both variants into one since they share similar colors and interactive properties. Additionally, it would be a good idea to include a brief description under the Color section to highlight their similar properties.

----

# Structure

#### Default toggle structure
- [x] Add the word "structure" at the end of the heading
- [x] Update the handle size in the image, as it is marked incorrectly (does not match with the table).

#### Small toggle structure
- [x] Add the word "structure" at the end of the heading
- [x] The table is missing a row for the Label text

---

### PR
- [x] Create PR
- [ ] PR review

Kritvi-bhatia17:
Closes #4216 

Update the Style tab with the latest content template and fixed a few more sections.

#### Changelog

**Changed**

- Converted all the "action text" to "state text", both in images and texts.
- Updated a few images to change "Label" to "Reminder".
- There are two color tokens in this table that do not have the $ before the token name, but they need it.
- In the default toggle structure table, updated the handle size in the image, as it is marked incorrectly (does not match with the table).
- In the small toggle structure table, added a row for the Label text.
- The image caption should not have a period after it.
- Fixed a few typos


src/pages/components/toggle/style.mdx
```diff
@@ -38,31 +41,31 @@ structure, and size.
...
-![On and off toggle states](images/toggle-style-color.png)
+![On and off states of default and small toggle](images/toggle-style-color.png)
 
 </Column>
 </Row>
 
-<Caption>On and off toggle states</Caption>
+<Caption>On and off states of default and small toggle</Caption>
 
 ### Interactive states
```
laurenmrice, line 51:
```suggestion
### Interactive state color
```

src/pages/components/toggle/style.mdx
```diff
@@ -72,34 +75,34 @@ and read-only states.
 <Row>
 <Column colLg={8}>
 
-![Interactive toggle states](images/toggle-style-states.png)
+![Example of default and small toggle interactive state color](images/toggle-style-states.png)
```
laurenmrice, line 78: Small thing, but in this interactive state color image, the labels for the focus state should say "Focus" and not "Focused", just to match the table terminology.

src/pages/components/toggle/style.mdx
```diff
@@ -143,7 +147,7 @@ phrase and any proper nouns capitalized, and no more than three words.
 
 </div>
 
-<Caption>Toggle sizes | px / rem</Caption>
+<Caption>Default, and small sizes of toggle | px / rem</Caption>
```
laurenmrice, line 150:
```suggestion
<Caption>Default and small sizes of toggle | px / rem</Caption>
```

laurenmrice (changes requested):
**Size section**
The column header for Height should read: "Height (px/rem)" according to the content template.

emyarod (approved):
looks good to me!

redeploying the vercel build

thyhmdo (approved):
looks good

alina-jacob (approved):
LGTM! 🚀

Kritvi-bhatia17:
Hi dev! Could you please help with merging the PR? Thanks!

laurenmrice (approved):

PR #4448, opened 2025-01-28
[Tooltip] Update Usage tab

### LINKED ISSUE #3956: Toggletip vs. Tooltip Docs Guidance Update
## Acceptance criteria

- We need to do some research on how we want to structure Tooltip and its variants versus Toggletip.
- There have been discussions about removing the "Standard" Tooltip variant, but also concern if people actually need that use case still and use it in code. We need to first see if we can find valid use cases for this before making the decision of removing it.
  - Ref: [a11y Tooltip ombnibus issue](https://github.com/carbon-design-system/carbon-website/issues/3332)
  - Ref: [a11y Toggletip omnibus issue](https://github.com/carbon-design-system/carbon-website/issues/3334)
- There were also discussions from a11y about only using the information icon in Toggletips and not at all in Standard tooltips, is this valid or are there instances where using the information icon could be used in Tooltips?

### Tasks
- [x] Playback findings to design and a11y team
- [x] Take feedback from playback and align on one direction to move forward with for Toggletip and Tooltip variants
- [ ] Update the Default Tooltip visual in the Storybook

### LINKED ISSUE #4441: [Tooltip] Doc update on Usage tab
Currently, we can't use IconButton or Button hasIconOnly because they use Tooltip internally and the story code wouldn't actually show how to use a Tooltip.

[Slack thread](https://ibm-studios.slack.com/archives/GSCEH4RSP/p1733935377188619)

-----

### 01/27/25

Discussed with Mike Gower and Taylor Jones. Tooltips should be integrated with other components (icon button, tab with icons, copy button,...) in the code. If not, there should be an issue to fix these bugs. This leaves only DefinitionTooltip as one variant of Tooltip. Mike suggested that the tooltip could be a pattern as well. 

----

**Usage tab**

- [x] De-emphasize Standard and Icon button Tooltips while leaving Definition Tooltip as one variant in this template
- [x] Make sure to pair documentation about tooltip of other components on the storybook 
- [x] Leaving images for IconButton tooltip is ok
- [x] Update some outdated info (interactions for definition tooltip)

thyhmdo:
Closes #4441 

- Tooltip Usage tab only
- Clarified the usage of the tooltip
- Only focused on clarifying tooltips that should be already integrated with other components. 

---

**Changed**

- Updated this to be the one variant template

**Removed**

- Icon button 
- Standard tooltip


alisonjoseph (approved):

src/pages/components/tooltip/usage.mdx
```diff
@@ -16,16 +16,22 @@ clarity to a user.
 
 </PageDescription>
 
+<InlineNotification>
+
+**New update** We have removed standard tooltip and icon button tooltip as
+variants from this guidance. There’s no need to include a separate tooltip
+within the component. There are no breaking changes, but we are working on
```
aagonzales, line 23: I also don't fully understand what you mean about it being within other components.

aagonzales (commented):
I noticed the definition tooltip was missing from the live demo. It maybe helpful to get that added in so people can see what the remaining variants are.

src/pages/components/tooltip/usage.mdx
```diff
@@ -16,16 +16,23 @@ clarity to a user.
 
 </PageDescription>
 
+<InlineNotification>
+
+**Documentation update** We've updated the guidance to no longer call out
```
aagonzales, line 21: Just need a colon here I think
```suggestion
**Documentation update:** We've updated the guidance to no longer call out
```

src/pages/components/tooltip/usage.mdx
```diff
@@ -57,11 +64,12 @@ clarity to a user.
 
 ## Overview
 
-A tooltip is a message box displayed when a user hovers over or focuses on a UI
-element. A tooltip is used to provide more information and should be paired with
-an interactive UI element like a button. Tooltips should be used sparingly and
-contain succinct supplementary information. Do not include interactive elements
-in tooltips; use the toggletip component instead.
+A tolltip is embedded within other components rather than being used as a
```
aagonzales, line 67:
```suggestion
A tooltip is embedded within other components rather than being used as a
```

src/pages/components/tooltip/usage.mdx
```diff
@@ -71,27 +79,43 @@ in tooltips; use the toggletip component instead.
...
-- To expose names of controls, like icon buttons, that lack visual labels
-- When an element can take focus and supplying additional information is useful
-  in helping a user make decisions
-- When an element needs more context or explanation
-- Use when defining a term or inline item
+By default, most components such as icon buttons, icon tabs, sliders, and copy
+buttons already have built-in tooltips, so there’s no need to add one
+separately. Customization is allowed but not recommended.
+
+Tooltips can be used to:
```
aagonzales, line 88: Should this be an H4? Could go either way.

aagonzales (approved):
This looks good Thy. Just two small typos. I'm going to go ahead and approve it though.

src/pages/components/tooltip/usage.mdx
thyhmdo: Yeah, the intention is that it's still connected to the first paragraph so I'd leave it as the body text.

PR #4477, opened 2025-02-19
doc(list): Update the Style tab with the latest content template

### LINKED ISSUE #4424: List: Style tab
## Acceptance criteria

Update the List component Style tab documentation with the new [content template](https://carbondesignsystem.com/contributing/documentation/#style-template:-for-components-with-multiple-variants).

------

# Description
- [x] Add this section.

# Anchor links
- [x] Add anchor links.

------

> Notes from the Playback:
> It’s fine to combine both variants into one since they share similar colors and interactive properties. Additionally, it would be a good idea to include a brief description under the Color section to highlight their similar properties.

---

# Color

## Unordered and ordered list color
- Add an H3 heading under "color" and rename it to this.
- [x] Image: Start with Unordered first, then Ordered and revise the caption. We will be talking about lists in that order on the Usage tab, so trying to keep that order consistent here as well.

------

# Structure
- [x] Image: Start with Unordered first, then Ordered and revise the caption. We will be talking about lists in that order on the Usage tab, so trying to keep that order consistent here as well.
- [x] Change specs to use Sans and not Mono.

------

# Feedback
- [x] Add this section.

---

### PR
- [x] Create PR
- [x] PR review

Kritvi-bhatia17:
Closes #4424 

Update the Style tab with the latest content template and fixed a few more sections.

#### Changelog

**New**

- Added a few sections:
   - Description 
   - Anchor links
   - Feedback

**Changed**

- Updated the order of the list's variants in all sections (text & image): Start with Unordered first, then Ordered
- Updated a few images
- Updated "dash" to "en dash"


src/pages/components/list/style.mdx
```diff
@@ -36,9 +54,18 @@ use the larger expressive 16px options.
 
 ## Structure
 
-There are two types of lists, ordered and unordered. Level 1 ordered list points
-are marked numerically while level 2 points are marked with letters. Level 1
-unordered list points are dashed while level 2 points are marked with a dot.
+There are two types of lists: unordered and ordered. Both have the same
```
laurenmrice, line 57: For this section, instead of introducing more heading levels, I would just stick with being concise in one paragraph and let the image visual do the work to explain dashes, letters, etc. Also I would either use the word "marked" or "prefixed" when thinking about how to describe them. And think about how you will be describing this in the Anatomy section on the Usage tab to be consistent with the terminology you want to use. Here is an example below. 

Frame 1

laurenmrice (changes requested):

src/pages/components/list/style.mdx
Kritvi-bhatia17: Thanks @laurenmrice, confirmed with dev:
> "In the [spec](https://www.w3.org/TR/css-lists-3/#markers) they are formally referred to as Markers." 
So, I have updated accordingly.

laurenmrice:
@Kritvi-bhatia17 The only other thing I see is that the structure image in the figma file needs to be 736px wide instead of 608px because it uses the fixed image component. Then it should render properly.

Kritvi-bhatia17:
> @Kritvi-bhatia17 The only other thing I see is that the structure image in the figma file needs to be 736px wide instead of 608px because it uses the fixed image component. Then it should render properly.

Oh, I have fixed that @laurenmrice

laurenmrice (approved):

Kritvi-bhatia17:
Hi @kenny or @guidari, can you please review this PR, whenever you get some time. Thank you!

Gururajj77 (approved):

PR #4542, opened 2025-04-11
docs: add import to clarify dependencies

rodet:
The code can't be run without the use statement. I had to look up for the right one in the source code, so let's rather have it directly in the example so copy & paste just works.

#### Changelog

**New**

- Add the correct use statement in the example


alisonjoseph (approved):

Kritvi-bhatia17 (approved):
Thanks for fixing it @rodet! 🚀

PR #4548, opened 2025-04-16
docs: remove mention of idl playbacks on the meetups page

laurenmrice:
This PR removes a mention about joining IDL playbacks on the Meetups page, which we no longer host anymore. If at some point we decide to host this meeting again, we can add it back, but for now it is misleading information.

---

**Removed**

- Image, button, and content


Kritvi-bhatia17 (approved):
Good catchup! LGTM 🚀

PR #4606, opened 2025-05-29
docs: correct spelling issues throughout codebase

adamalston:
No issue.

Corrected spelling issues in the docs and elsewhere in the codebase.

#### Changelog

**Changed**

- Corrected spelling issues in the docs and elsewhere in the codebase.

---

I left a `TODO` with some questions that hopefully someone can answer.

### DIFF
--- a/src/pages/components/tabs/style.mdx
+++ b/src/pages/components/tabs/style.mdx
@@ -105,6 +105,18 @@ structure, and size.
   Examples of enabled, hover, and active states for scrollable line tabs.
 </Caption>
 
+{/* prettier-ignore-start */}
+
+{/* 
+TODO: 
+1. Are all of the 'dismissible' strings supposed to be 'dismissable' instead?
+The Carbon Storybook uses 'dismissable' (https://react.carbondesignsystem.com/?path=/docs/components-tabs--overview#dismissable-tabs)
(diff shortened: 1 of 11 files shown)


src/pages/components/tabs/style.mdx
```diff
@@ -105,6 +105,18 @@ structure, and size.
   Examples of enabled, hover, and active states for scrollable line tabs.
 </Caption>
 
+{/* prettier-ignore-start */}
+
+{/* 
+TODO: 
+1. Are all of the 'dismissible' strings supposed to be 'dismissable' instead?
```
kennylam, line 112: It should be spelled correctly, IMO. Not sure of the original intent (I'm guessing just a typo), but it's either gone unnoticed or just easier to leave it as-is. This would be good to fix for v12, I think. We could introduce the new (correctly spelled) one now with a deprecation notice for the old one.

Unless there's context I'm missing, @tay1orjones?

laurenmrice: We have been calling this `dismissible`, and not dismissable across the board for code and design. I would like to get input from the IBM Style council on what the correct spelling of this should be, because I think it could technically be either of these.

adamalston: Should I expect a reply here when you receive feedback or should I wait for it elsewhere? Not sure if an internal Slack discussion was started.

PR #4676, opened 2025-08-01
chore: Update docs, resolves #4520

### LINKED ISSUE #4520: [Docs]: The doc for applying a type style is hidden from the visible viewport
### Project

https://www.carbondesignsystem.com

### Section

https://carbondesignsystem.com/elements/typography/code/

### Problem

Very often I see people who don't know how to style fonts with Carbon. They use hardcoded weight, style etc, and invariably fail at applying the right styles. This results in poor typography, inconsistent products, bad readability and more.

Image

I just found out that the usage section of the typography page mislead me for a good while. The actual way to set the font for a specific class is hidden from the default view. Looking at the box I thought it was only about setting the includes right. I was surprised when, wanting to copy this code, another super important information was hidden below the "fold".

Image

### Solution

I suggest pulling the section `.selector { [...] }` into a new paragraph so it is easy to find:

> ### How to style an element
>
> Outside of very conscious exceptions, style all your fonts with the Carbon helpers so you can easily
> keep your code robust and consistent. Avoid using single CSS attributes like `font-weight` or `font-size`
> whenever possible and rather defer the type handling to Carbon.
>
> ```scss
> .selector {
>  // Include a type style
>  @include type.style('productive-heading-01');
> }

### Suggested Severity

None

### Application

_No response_

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

anjju:
Closes #4520

#### Changelog

**Changed**

- In https://carbondesignsystem.com/elements/typography/code/ pull the section .selector { [...] } into a new paragraph so it is easy to find:

```

How to style an element
Outside of very conscious exceptions, style all your fonts with the Carbon helpers so you can easily
keep your code robust and consistent. Avoid using single CSS attributes like font-weight or font-size
whenever possible and rather defer the type handling to Carbon.

.selector {
 // Include a type style
 @include type.style('productive-heading-01');
}
```

### DIFF
--- a/src/pages/elements/typography/code.mdx
+++ b/src/pages/elements/typography/code.mdx
@@ -47,12 +47,12 @@ You can use this package by writing the following:
 @include type.type-classes();
 ```
 
-## How to style an element
+### Type-styles
 
-Outside of very conscious exceptions, style all your fonts with the Carbon
-helpers so you can easily keep your code robust and consistent. Avoid using
-single CSS attributes like font-weight or font-size whenever possible and rather
-defer the type handling to Carbon.
+Unless there are specific exceptions, always use Carbon helpers to style your
+fonts. This ensures your code stays consistent and maintainable. Avoid directly
+setting individual CSS properties like font-weight or font-size; instead, rely
+on Carbon to handle all typography styling.
 
 ```scss
 .selector {
@@ -141,7 +141,7 @@ function:
 <Row className="resource-card-group">
   <Column colMd={4} colLg={4} noGutterSm>
     <ResourceCard
-      subTitle="Carbon type package"
+      subTitle="Carbon type package" 
       href="https://github.com/carbon-design-system/carbon/tree/main/packages/type"
     >
 


alina-jacob:
Thanks for the PR, looking good! @anjju ⭐ 

Just noting a couple of things, 
- I'd update the heading from "How to style an element" to "Type-styles"
- I'd change the heading "How to style an element" from H2 to H3
- I'd also rephrase the paragraph to "Unless there are specific exceptions, always use Carbon helpers to style your fonts. This ensures your code stays consistent and maintainable. Avoid directly setting individual CSS properties like font-weight or font-size; instead, rely on Carbon to handle all typography styling."

@thyhmdo, @laurenmrice, feel free to add any other recommendations!

PR #4780, opened 2026-01-06
fix(live demo): update live labels and add feature flag toggle

### LINKED ISSUE #4683: [Toggle] add the feature flag toggles to the live demo as options in the menu
## Acceptance criteria

- [ ] Add the feature flag toggles to the live demo as options in the menu.
- [ ] Name them something like: "`Default toggle (preview)`" and "`Small toggle (preview)`"

- [ ] Having said that the [Tile](https://carbondesignsystem.com/components/tile/usage/#live-demo) and [Structured list](https://carbondesignsystem.com/components/structured-list/usage/#live-demo) live demo names in the menu need to be updated from `(unstable)` to `(preview)` as well.

alisonjoseph:
Closes #4683

#### Changelog

**New**

- Add feature flag toggle to live demo https://react.carbondesignsystem.com/?path=/story/components-toggle-feature-flag--toggle

**Changed**

- Rename all instances of unstable in demos to preview


alisonjoseph:
@Kritvi-bhatia17 We would need to create two different stories in order to show in the live demo. So I suppose we would need to decide if thats something we want to do here. We don't really do this for most other components/props.

PR #4818, opened 2026-02-05
fix(modal-demo): increase live demo container size

### LINKED ISSUE #4806: [Bug]: Live demo of the modal component on website is very short
The live demo for the modal component on the https://carbondesignsystem.com/components/modal/usage/#live-demo web page is very short leading to a quite unrepresentative experience of how a modal should appear.

I'd suggest if possible the height of the live demo component be at least doubled to provide a reasonable height for the modal and its contents. This would give a more representative experience of how a modal should look, and how certain behaviours such as the scroll behaviour when a user clicks into a select field, or opens a dropdown. 

Currently it's 20REM which looks like this. 

Image

I'd propose it should be at least 40REM which would look like this

Image

maradwan26:
Closes #4806

Increases the size of the live demo in `/components/modal/usage/#live-demo`

#### Changelog

**New**

- Added `wide` prop to  the StorybookDemo

### DIFF
--- a/src/pages/components/modal/code.mdx
+++ b/src/pages/components/modal/code.mdx
@@ -81,6 +81,7 @@ considers all interactive child nodes and wraps focus based on their tab order.
   themeSelector
   url="https://react.carbondesignsystem.com"
   tall
+  wide
   variants={[
     {
       label: 'Default',


Kritvi-bhatia17 (approved):
LGTM 🚀

PR #4830, opened 2026-02-13
docs: Added images for Claude Desktop setup

### LINKED ISSUE #4834: [Docs]: Add images for Carbon MCP setup drop downs
### Project

https://www.carbondesignsystem.com

### Section

https://carbondesignsystem.com/developing/carbon-mcp/onboarding-and-setup/

### Problem

Currently the drop downs for Claude Desktop, Cursor, and VS Code set up is mostly text. 

### Solution

Update the document with hi-res screenshots, showing the setup step instead of just having text for them.

- [ ] Claude Desktop
- [ ] Cursor
- [ ] VS Code

### Suggested Severity

None

### Application

_No response_

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

ChanceG0:
Closes #4834 

Images for Claude Desktop setup

#### Changelog

**New**

- images in claude desktop drop down

**Changed**

- spacing inbetween images
- more specific text instructions

**Removed**

- older images


src/pages/developing/carbon-mcp/onboarding-and-setup.mdx
```diff
@@ -202,27 +202,40 @@ This section describes how to make both connections.
...
+  [here](./files/carbon-mcp-claude-extension.mcpb).
 
 <br />
+![Claude Desktop Step 3](./images/Claude-Desktop-"Install-the-Carbon-1".png)
+<br />
+![Claude Desktop Step 4](./images/Claude-Desktop-"Install-the-Carbon-2".png)
 <br />
 
-- You will see Carbon MCP configured under installed extensions
+- Enter your Carbon MCP auth token and session ID in the modal
```
Kritvi-bhatia17: How about combining lines 226 and 227 like this?

```suggestion
- Enter your Carbon MCP auth token and session ID in the modal, and click Save to confirm.
```

Kritvi-bhatia17 (commented):
Looks good @ChanceG0, thanks for creating the issue as well! 🚀

Hey @ariellalgilmore, quick question — is there a way to make the text expand to fill the full width? I’ve included an image below for reference.


Image

Chance-Gong:
@Kritvi-bhatia17 I looked into this, and the reason it's not wrapping the full width is the `-` for the list. Removing the dashed list fixes the wrapping issue, but also removes the - before the step. 

I also noticed the Bob client drop-down doesn't have the dashes, and the text wrapping is fine.
Perhaps we could remove the dashes for all other steps as well?

Screenshot 2026-02-18 at 2 20 48 PM

Screenshot 2026-02-18 at 2 33 50 PM

Kritvi-bhatia17:
> @Kritvi-bhatia17 I looked into this, and the reason it's not wrapping the full width is the `-` for the list. Removing the dashed list fixes the wrapping issue, but also removes the - before the step.
> 
> I also noticed the Bob client drop-down doesn't have the dashes, and the text wrapping is fine. Perhaps we could remove the dashes for all other steps as well?


Thanks for looking into it, @Chance-Gong. In that case, it might feel inconsistent with the section below, which uses bullet points. Ideally, we should adjust it so the content expands to the full width with bullets, if that's possible.

PR #4970, opened 2026-09-02
docs (Carbon MCP): Updates for GA documentation

sct-will:
## Motivation

Carbon MCP is moving to GA. The docs still described the public-preview,
token-based setup and could be streamlined. This updates them for OAuth
discovery and the carbon-builder skill, and trims each page.

## Changes

**Overview**
- Removed the "Public Preview" label and all preview / early-access content.
- Combined "Model Context Protocol" and "Why Carbon MCP" into a single lead
  sentence and dropped the anchor links.
- Replaced the empty onboarding tile with a button, added the `labs_search` and
  `code_audit` tools (list alphabetized), and moved content coverage and roadmap
  into accordions.

**Onboarding and setup**
- Reframed Step 2 around OAuth: most clients add the server URL and sign in, with
  token + session ID kept as a documented fallback for Bob, headless agents, and
  older clients. Removed the request-access flow.
- Moved Cursor, Claude Code, and VS Code to OAuth; added Windows (PowerShell)
  install commands alongside macOS/Linux.
- Flattened the nested accordions into a single collapsed "Other clients" per
  step, promoted Figma Make to the top set, and fixed the heading hierarchy.
- Repointed skill downloads to the GitHub zip; removed the orphaned local
  `carbon-builder.zip` and the obsolete Claude Desktop `.mcpb` extension.

**Prompts**
- Rewrote the DOs and DON'Ts around the carbon-builder skill and added "ask
  clarifying questions and plan first."
- Removed the long prompt-starter template and replaced three long samples with
  five short ones, one per MCP tool.
- Trimmed the Overview into the lead and tightened the copy throughout.


sct-will:
This content will also be added to the Payload instance. Landing the change here first.

src/pages/developing/carbon-mcp/overview.mdx
```diff
@@ -6,107 +6,54 @@ description:
...
-
-## Why Carbon MCP?
+<div style={{ marginTop: '0.5rem', marginBottom: '2.5rem' }}>
+  <Button
+    kind="tertiary"
+    href="/developing/carbon-mcp/onboarding-and-setup/"
+    renderIcon={ArrowRight}>
+    {'Onboarding and setup'}
+  </Button>
+</div>
```
elycheea, line 30: So this actually won’t translate super well in PayloadCMS since our page descriptions will move into the page header.

In PayloadCMS, we followed a strict content structure where content needs to be within a section with a heading. The video won’t be able to standalone.

| Deploy preview | In PayloadCMS | Proposed changes for PayloadCMS
-----------------|---------------|-----------------------
Image | Image | Image

I also can’t think of (m)any places where we use CTAs within content as you see here? _Maybe_ a `MiniCard` could work, but we’ve also been trying to limit its use in case we decide to deprecate it.

(Also this is being tested on my local where I’m able to generate a new collection for Carbon MCP to test IA suggestions. The cards are just leftover copying the content while trying to demonstrate the video issue.)

src/pages/developing/carbon-mcp/overview.mdx
```diff
@@ -127,11 +74,19 @@ Labs UIShell.
...
+      download>
+      <MdxIcon name="carbon" />
+    </ResourceCard>
+  </Column>
 </Row>
 
-## Summary of content coverage
-
-Details about current coverage is provided below:
+<Accordion>
```
elycheea, line 88: For PayloadCMS, I would honestly recommend against collapsing **Content coverage** and **Roadmap** into accordions since you’d lose them in the table of contents.

Image

src/pages/developing/carbon-mcp/overview.mdx
```diff
@@ -191,3 +141,6 @@ coverage, and optimizing the experience based on user feedback.
 - Carbon Patterns context
```
elycheea, line 141: Interestingly, I see Carbon Labs listed in the Payload version. I don’t see the coverage adding in current though even though it was presumably removed here?

src/pages/developing/carbon-mcp/overview.mdx
```diff
@@ -174,14 +129,9 @@ Details about current coverage is provided below:
...
-#### <CheckmarkFilled size={16} color={green[40]} /> Carbon Labs (`@carbon/labs-react`)
+</AccordionItem>
 
-- Code examples and documentation
-- Example: AnimatedHeader, Processing, Resizer, WhatsNew, Labs UIShell
-- Full code examples
-- Complete props and imports
-
-## Roadmap
+<AccordionItem title="Roadmap">
```
elycheea, line 134: Probably more of a content question for our content folks, but we’ve generally shied away from including roadmaps directly since they become outdated very easily on the websites. Instead, it might be worth linking to a GitHub board if you maintain a roadmap separately there.

src/pages/developing/carbon-mcp/onboarding-and-setup.mdx
elycheea, line 1: This is a pretty tricky section since we’re looking for ways to reduce duplication of content while keeping it easily accessible.

Here’s one of my earlier attempts.

Image

The main changes in this proposal are:
1. **Move step 1 (IBMid / w3id) into _part_ of the setup step.** The reasoning for this is that this is more of a prerequisite than a part of the actual setup.<br/><br/> We do something similar (although wordier) in the [Instrumentation docs](https://preview.carbondesignsystem.com/getting-started/developing/instrumentation/overview).

2. **Distinguish OAuth v. token approach for setup.** I’m reading 2-3 recommended approaches in here — using the server URL OR the JSON provided if OAuth is supporting, if not, using token and session ID.

I’m also open to exploring this with steps following by a set or resources/mini cards, similar to the Instrumentation docs too.

elycheea: @oliviaflory and I are both exploring options for step 2 (MCP clients).

One main theme is to move IBM Bob, Claude, Figma Make, Other clients into individual tabs. One benefit to tabs, is that for the Claude tab, space is less of a constraint and we _could_ cover Claude Code and Desktop again. The individual guidance for `carbon-builder` also fits better when we break these into separate tabs.

elycheea (commented):
Documenting some initial feedback, especially as we think about how this content moves into PayloadCMS.

The onboarding section definitely feels a bit tricky — @oliviaflory and I have both been testing a few options that might work better when it moves to Payload.

Initial scan of prompts looks okay but will give it a closer look tomorrow!

